PR #1, opened 2017-01-03
Accumulo 4510

mikewalch:
To make this review easier, I pushed the commit ac5b271caec45083751efad7298852969bec387e to master that copied the code from Accumulo.  The commit has some changes to package names and POM file so feel free to review.  

This PR contains code that improves how random walk tests are run.  Users can now run them locally or in YARN.


README.md
```diff
@@ -1 +1,50 @@
...
+
+The command below will start a local random walker using the [Image.xml][image] module.
+
+        ./bin/accumulo-testing rw-local Image.xml
+
+To run multiple random walkers on a cluster using YARN, set `test.container.instances` in
+`accumulo-testing.props` to the number of walkers that you want started and run the command
+below.
+
+        ./bin/accumulo-testing rw-cluster Image.xml
```
joshelser: I'd suggest `rw-yarn` or `rw-twill` instead of `rw-cluster` for some future proofing.

README.md
```diff
@@ -1 +1,50 @@
...
+
+3. Tests are run using the `accumulo-testing` command which is located in the `bin/`
+   directory. Run this command without any arguments to view its usage and see available tests.
+
+        ./bin/accumulo-testing
+
+## Random walk test
+
+The random walk test generates client behavior on an Apache Accumulo instance by randomly walking a
+graph of client operations. The test is configured by the `accumulo-testing.props` file and requires
```
joshelser: Define here what the minimum set of properties are? Also, would prefer the standard "properties" suffix, but that's just me being picky.

bin/accumulo-testing
```diff
@@ -0,0 +1,89 @@
...
+EOF
+}
+
+if [ -f $AT_HOME/conf/accumulo-testing-env.sh ]; then
+  . $AT_HOME/conf/accumulo-testing-env.sh
+else
+  . $AT_HOME/conf/accumulo-testing-env.sh.example
+fi
+
+export AT_PROPS=$AT_HOME/conf/accumulo-testing.props
```
joshelser: shellcheck: quotes needed.

bin/accumulo-testing
```diff
@@ -0,0 +1,89 @@
...
+  java -Dlog4j.configuration=file:$log4j_config -cp $AT_SHADED_JAR $randomwalk_main $AT_PROPS $2
+  ;;
+rw-cluster)
+  if [ -z $2 ]; then
+    echo "ERROR: <module> needs to be set"
+    print_usage
+    exit 1
+  fi
+  build_shade_jar
+  mvn compile -Ptest-runner -Dhadoop.version=$HADOOP_VERSION -Dexec.args="-n AccumuloRandomWalkTest -j $AT_SHADED_JAR -m $randomwalk_main -t $AT_PROPS -l $log4j_config -a ./accumulo-testing.props $2"
```
joshelser: compile? Not package?

conf/accumulo-testing-env.sh.example
```diff
@@ -0,0 +1,25 @@
...
+# Hadoop
+# ======
+test -z "$HADOOP_PREFIX" && export HADOOP_PREFIX=/path/to/hadoop
+test -z "$HADOOP_CONF_DIR" && export HADOOP_CONF_DIR=/path/to/hadoop/etc/hadoop
+
+# Accumulo
+# ========
+# Set the Accumulo version that should be included in the shaded jar
+export ACCUMULO_VERSION=`accumulo version`
+export HADOOP_VERSION=`hadoop version | grep Hadoop | cut -d ' ' -f 2`
```
joshelser: I think `hadoop version | head -n1 | awk '{print $2}'` is a little more clear here.

twill/src/main/java/org/apache/accumulo/testing/twill/TestRunner.java
```diff
@@ -0,0 +1,166 @@
...
+    }));
+
+    LOG.info("Waiting for {} to finish in YARN...", opts.testName);
+    LOG.info("Press ctrl-c to kill {} in YARN", opts.testName);
+
+    try {
+      controller.awaitTerminated();
+      done.set(true);
+    } catch (ExecutionException e) {
+      LOG.error("Error", e);
```
joshelser: Wouldn't we want to re-throw the exception so that we get a non-zero exit code?

twill/src/main/java/org/apache/accumulo/testing/twill/TestRunner.java
```diff
@@ -0,0 +1,166 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class TestRunner {
```
joshelser: More descriptive class name, please.

twill/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+  <packaging>jar</packaging>
+
+  <name>Apache Accumulo Testing Twill</name>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>13.0.1</version>
```
joshelser: Why not have this defined in `accumulo-testing`?

bin/accumulo-testing
```diff
@@ -0,0 +1,89 @@
...
+fi
+
+function build_shade_jar() {
+  export AT_SHADED_JAR=$AT_HOME/core/target/accumulo-testing-core-$AT_VERSION-shaded.jar
+  if [ ! -f $AT_SHADED_JAR ]; then
+    echo "Building $AT_SHADED_JAR"
+    cd $AT_HOME
+
+    : ${ACCUMULO_VERSION?"ACCUMULO_VERSION must be set in bash env or conf/accumulo-testing-env.sh"}
+    mvn clean package -Pcreate-shade-jar -DskipTests -Daccumulo.version=$ACCUMULO_VERSION -Dhadoop.version=$HADOOP_VERSION
```
joshelser: Tricky ;)

We should be making sure that the jar which is created is properly licensed (that is un-skirtable). Have you taken a look to see if it looks "sane"?

bin/accumulo-testing
```diff
@@ -0,0 +1,89 @@
...
+  fi
+fi
+
+function build_shade_jar() {
+  export AT_SHADED_JAR=$AT_HOME/core/target/accumulo-testing-core-$AT_VERSION-shaded.jar
+  if [ ! -f $AT_SHADED_JAR ]; then
+    echo "Building $AT_SHADED_JAR"
+    cd $AT_HOME
+
+    : ${ACCUMULO_VERSION?"ACCUMULO_VERSION must be set in bash env or conf/accumulo-testing-env.sh"}
```
joshelser: I actually have no clue what this colon is doing here. Is this a typo or just some weird bash-ism I'm not familiar with?

README.md
mikewalch: Yes min set of properties is needed and I will change suffix to standard.

README.md
mikewalch: `rw-yarn` sounds good to me.  I will change.

bin/accumulo-testing
mikewalch: will fix

twill/pom.xml
mikewalch: The guava version in the main pom is set to `14.0.1` which coincides with Accumulo.  Twill has a hard requirement on `13.0.1`.  Things break even if `14.0.1` is used.  I tried this and the version can be moved down to where guava is listed in dependencies.  I will also add comment.

bin/accumulo-testing
mikewalch: I took a look at and it looks sane to me.

bin/accumulo-testing
joshelser: OK. Definitely don't change the bash if it works just because I don't recognize it. If you could slap in a comment as to what it's doing, that's more than enough for me :)

bin/accumulo-testing
joshelser: Hm. @ctubbsii knows off the top of his head if this would cause any problems? Maybe `exec:java` can use the compiled class files (and doesn't require a jar).

joshelser:
Thanks, Mike! I'll try to take another look, but feel free to push now if you'd like. I trust your judgement.

bin/accumulo-testing
mikewalch: no worries.  i changed it.  the check is higher up in the code now after accumulo-testing-env.sh is sourced.

bin/accumulo-testing
mikewalch: See yarn/pom.xml for the exec-maven-plugin config.  I cannot just use `mvn exec:java` on the command line as the class that I want to execute is in a module.  The execution is being attached to the compile phase as a jar isn't required.

ctubbsii: The `build_shade_jar` function does a `mvn package`, and the `mvn compile` here is just to trigger the `exec:java` from the compiled code. I'm sure there's a more efficient way to do this, because this seems a bit redundant, but I can't see any immediate problems with it.

joshelser: Thanks for the explanation, Mike and Christopher. +1

conf/accumulo-testing-env.sh.example
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+# Hadoop
+# ======
+test -z "$HADOOP_PREFIX" && export HADOOP_PREFIX=/path/to/hadoop
+test -z "$HADOOP_CONF_DIR" && export HADOOP_CONF_DIR=/path/to/hadoop/etc/hadoop
+
+# Accumulo
+# ========
+# Set the Accumulo version that should be included in the shaded jar
```
keith-turner: would be nice to make the version of zookeeper included in the shaded jar configurable also

README.md
```diff
@@ -1 +1,54 @@
...
+        ./bin/accumulo-testing rw-local Image.xml
+
+If you would like to run multiple, distributed random walkers, run the command below to start random
+walkers in 5 containers in YARN using the Image.xml module.
+
+        ./bin/accumulo-testing rw-yarn 5 Image.xml
+
+This command will create an application in YARN and exit when test is completed. While its
+running, you can view logs for each random walker using the YARN resource manager. The YARN
+application can be killed at any time using `ctrl-c` or via the resource manager.
```
keith-turner: If you kill the application driving the test using `ctrl-c`, will that cause things in yarn to die?

mikewalch: yes it will.

keith-turner: So can I ssh to a cluster, start random walk on N node, and then log off?  Maybe I need to run the application that stated the job with nohup.

conf/accumulo-testing-env.sh.example
mikewalch: this added in 45b9dbfbbca

mikewalch:
I think that this pull request is ready to merged. Before this is merged, I want to mention two more commits that were added to PR in case any one wants to review:

* 45b9dbfbbcaf - I fixed a bug that was preventing RW tests from launching MapReduce jobs.  I also removed test code that depended on Accumulo being installed locally.
* 36f31026561de99973e - I changed the behaviour of the `accumulo-testing rw-yarn` command to exit after all containers of the YARN application have been created but before test has ended.

PR #3, opened 2017-01-30
ACCUMULO-4579 Fixed hadoop config bug in accumulo-testing

mikewalch:
* TestEnv now returns Hadoop configuration that is loaded from config files
  but expects HADOOP_PREFIX to be set in environment
* Fixed bug where Twill was being improperly configured to look in wrong
  location for shaded jar when running test application in YARN.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/testing/core/TestEnv.java
+++ b/core/src/main/java/org/apache/accumulo/testing/core/TestEnv.java
@@ -19,7 +19,7 @@ import org.apache.accumulo.core.client.security.tokens.AuthenticationToken;
 import org.apache.accumulo.core.client.security.tokens.KerberosToken;
 import org.apache.accumulo.core.client.security.tokens.PasswordToken;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.mapreduce.tools.CLI;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.security.UserGroupInformation;
 
 public class TestEnv {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/accumulo/testing/core/TestEnv.java
```diff
@@ -96,15 +97,22 @@ public String getPid() {
...
-    config.set("fs.hdfs.impl", org.apache.hadoop.hdfs.DistributedFileSystem.class.getName());
-    config.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
-    return config;
+    if (hadoopConfig == null) {
+      String hadoopPrefix = System.getenv("HADOOP_PREFIX");
+      if (hadoopPrefix == null || hadoopPrefix.isEmpty()) {
+        throw new IllegalArgumentException("HADOOP_PREFIX must be sent in env");
+      }
+      hadoopConfig = new Configuration();
+      hadoopConfig.addResource(new Path(hadoopPrefix + "/etc/hadoop/core-site.xml"));
```
mikewalch: After talking to @keith-turner offline, I am going to submit an update that pull the needed Hadoop configuration from the accumulo-testing.properties file.

ctubbsii: I think using the properties file is better. I don't think we should be writing java code to depend on env variables, especially arbitrary script conventions like `HADOOP_PREFIX`, which is very sensitive to Hadoop packaging/deployment changes.

mikewalch: I pushed another commit 62e91527 where I created new properties in `accumulo-testing.properties` to avoid relying on loading Hadoop config files using `HADOOP_PREFIX`.  I also added `accumulo` to several properties that configured Accumulo scanners.

PR #6, opened 2017-10-12
ACCUMULO-4718 Add CLASSPATH to accumulo-testing script

milleruntime:
The test "rw-local Bulk.xml" was throwing errors because it couldn't find the HDFS containing the files to import.  The hadoop conf wasn't on the classpath so it was defaulting to the disk "file:///" file system.  This doesn't seem to affect the rest of the tests.  Not sure if this is the best place to include HADOOP_CONF_DIR...

### DIFF
--- a/bin/accumulo-testing
+++ b/bin/accumulo-testing
@@ -70,6 +70,7 @@ fi
 
 function build_shade_jar() {
   export at_shaded_jar="$at_home/core/target/accumulo-testing-core-$at_version-shaded.jar"
+  export CLASSPATH="$at_shaded_jar:$CLASSPATH"
   if [ ! -f "$at_shaded_jar" ]; then
     echo "Building $at_shaded_jar"
     cd "$at_home" || exit 1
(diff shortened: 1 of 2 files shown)


bin/accumulo-testing
```diff
@@ -229,7 +229,7 @@ rw-local)
     exit 1
   fi
   build_shade_jar
-  java -Dlog4j.configuration="file:$log4j_config" -cp "$at_shaded_jar" "$randomwalk_main" "$at_props" "$2"
+  java -Dlog4j.configuration="file:$log4j_config" -cp "$at_shaded_jar:$HADOOP_CONF_DIR" "$randomwalk_main" "$at_props" "$2"
```
ctubbsii: One way we can decouple this from specific environment variables (which may or may not exist), is to simply remove the `-cp` section, and set `CLASSPATH` above the call to `java`.
Something like `export CLASSPATH="$at_shaded_jar:$CLASSPATH"`

That way, the caller can simply set `CLASSPATH=/path/to/hadoop/confdir` prior to running the `accumulo-testing` command.

We could also add a sanity check prior to running java to make sure that the `$CLASSPATH` contains a directory with `core-site.xml` and/or `hdfs-site.xml` (or whatever the Hadoop client configuration uses by default).

milleruntime:
Sounds good.  Thanks

milleruntime:
@mikewalch As far as I can tell, I believe my changes in 52baec5 will also satisfy ACCUMULO-4723

It seemed to run fine with the quick test I ran using: accumulo-testing rw-yarn 1 Bulk.xml

PR #7, opened 2017-10-18
ACCUMULO-4717 Use public API

milleruntime:
* Replace security operations with API
* Utilize newer RFile API
* Replace core.client.impl code
* Replace CachedConfiguration with Hadoop Configuration

### DIFF
--- a/core/src/main/java/org/apache/accumulo/testing/core/randomwalk/concurrent/Config.java
+++ b/core/src/main/java/org/apache/accumulo/testing/core/randomwalk/concurrent/Config.java
@@ -20,8 +20,8 @@ import java.util.Properties;
 import java.util.SortedSet;
 
 import org.apache.accumulo.core.client.AccumuloException;
-import org.apache.accumulo.core.client.NamespaceNotFoundException;
-import org.apache.accumulo.core.client.TableNotFoundException;
+import org.apache.accumulo.core.client.impl.thrift.TableOperationExceptionType;
+import org.apache.accumulo.core.client.impl.thrift.ThriftTableOperationException;
 import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.testing.core.randomwalk.RandWalkEnv;
 import org.apache.accumulo.testing.core.randomwalk.State;
@@ -129,8 +129,10 @@ public class Config extends Test {
         try {
           env.getAccumuloConnector().tableOperations().setProperty(table, property.getKey(), property.getDefaultValue());
         } catch (AccumuloException ex) {
-          if (ex.getCause() instanceof TableNotFoundException) {
-            return;
+          if (ex.getCause() instanceof ThriftTableOperationException) {
+            ThriftTableOperationException ttoe = (ThriftTableOperationException) ex.getCause();
+            if (ttoe.type == TableOperationExceptionType.NOTFOUND)
+              return;
           }
           throw ex;
         }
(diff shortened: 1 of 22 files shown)


core/src/main/java/org/apache/accumulo/testing/core/randomwalk/shard/BulkInsert.java
```diff
@@ -172,7 +171,7 @@ private void sort(State state, RandWalkEnv env, FileSystem fs, String tableName,
 
     Collection<Text> splits = conn.tableOperations().listSplits(tableName, maxSplits);
     for (Text split : splits)
-      out.println(Base64.getEncoder().encodeToString(TextUtil.getBytes(split)));
+      out.println(Base64.getEncoder().encodeToString(split.getBytes()));
```
keith-turner: Be careful here.  Text has length.  The getBytes() method may return the internal byte array, which may be longer than the length.  I think this is the reason that TextUtil.getBytes() exists.

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/image/ImageFixture.java
```diff
@@ -64,15 +63,15 @@ public void setUp(State state, RandWalkEnv env) throws Exception {
     try {
       conn.tableOperations().create(imageTableName);
       conn.tableOperations().addSplits(imageTableName, splits);
-      log.debug("Created table " + imageTableName + " (id:" + Tables.getNameToIdMap(instance).get(imageTableName) + ")");
```
keith-turner: I think there are public apis to get the table id

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/concurrent/Config.java
```diff
@@ -190,10 +186,8 @@ private void changeTableSetting(RandomDataGenerator random, State state, RandWal
     try {
       env.getAccumuloConnector().tableOperations().setProperty(table, setting.property.getKey(), "" + newValue);
     } catch (AccumuloException ex) {
-      if (ex.getCause() instanceof ThriftTableOperationException) {
```
keith-turner: How did you find what exception to replace the cause with?

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/image/ImageFixture.java
milleruntime: Yeah there is, I can use it.

milleruntime:
Sorry I rebase'd a fix I found while running randomwalk Security.xml.

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/shard/BulkInsert.java
milleruntime: This case and a few others made me wonder why code from util is not part of the API?

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/concurrent/Config.java
milleruntime: Assuming this is the correct spot, I followed what the setProperty() method in TableOperationsImpl was doing... https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java#L897

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/shard/BulkInsert.java
ctubbsii: Well, ideally, Text will not be part of the API, eventually. It might make sense to move a few helpful utils into the public API in the interim.

keith-turner: Text has a copyBytes() method.  Would probably be best to use that here.  I think TextUtil method only copies if needed.

PR #8, opened 2017-10-19
ACCUMULO-4725 show rw modules dir in accumulo-testing

milleruntime:


bin/accumulo-testing
```diff
@@ -223,6 +224,12 @@ ci-mapred)
   build_shade_jar
   "$HADOOP_PREFIX"/bin/yarn jar "$at_shaded_jar" "$ci_main" "$at_props"
   ;;
+rw-list)
+  echo "Available randomwalk modules:"
+  for x in $at_home/core/src/main/resources/randomwalk/modules/*.xml ; do
```
keith-turner: For runtime, does mvn package put this in a diff place?

milleruntime: The module class looks on the classpath for these files.

milleruntime:
After some consideration, I will just add the path to the README for the modules.  This change to the script doesn't help if you need to edit the files or want to add new ones.

milleruntime:
It was already in the README so I added it to the help script since it wasn't super obvious.

PR #9, opened 2017-10-20
ACCUMULO-4717 Refactor WalkingSecurity to use API

milleruntime:
Simplified WalkingSecurity by removing the inheritance from server
classes and removing unused code. Now it is simply a helper class and
the other classes in the randomwalk framework call the API directly.


core/src/main/java/org/apache/accumulo/testing/core/randomwalk/security/Validate.java
```diff
@@ -101,12 +101,6 @@ public static void validate(State state, RandWalkEnv env, Logger log) throws Exc
       }
 
     }
-
-    Authorizations accuAuths = conn.securityOperations().getUserAuthorizations(WalkingSecurity.get(state, env).getTabUserName());
-    Authorizations auths = WalkingSecurity.get(state, env).getUserAuthorizations(WalkingSecurity.get(state, env).getTabCredentials());
-
-    if (!auths.equals(accuAuths))
-      throw new AccumuloException("Table User authorizations out of sync");
```
milleruntime: After removing the inheritance from WalkingSecurity, the call here to get auths was now an error.  It then became obvious that these 2 calls to getUserAuthorizations were doing the same thing.

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/security/AlterTable.java
```diff
@@ -34,13 +35,26 @@
...
-    boolean hasPermission = conn.securityOperations().hasTablePermission(WalkingSecurity.get(state, env).getSysUserName(), tableName,
-        TablePermission.ALTER_TABLE);
+    boolean hasPermission;
+    try {
+      hasPermission = conn.securityOperations().hasTablePermission(systemUser, tableName, TablePermission.ALTER_TABLE)
+          || conn.securityOperations().hasSystemPermission(systemUser, SystemPermission.ALTER_TABLE);
+    } catch (AccumuloSecurityException ae) {
+      if (ae.getSecurityErrorCode().equals(SecurityErrorCode.TABLE_DOESNT_EXIST)) {
+        if (exists)
+          throw new TableExistsException(null, tableName, "Got a TableNotFoundException but it should exist", ae);
```
keith-turner: why throw TableExistsExcep?

milleruntime: Because I copied what was used below.  Think they should all just be AccumuloException?

PR #11, opened 2017-12-15
ACCUMULO-4749 WIP Bulk Loading Test

jkrdev:
This is in no way ready to be merged, I just wanted to push the work I have done so far in hopes of having some eyes look over it and possibly give suggestions/advice. There are 3 classes that are empty the generator, loader, and ingest. Ingest will be deleted but at the moment I wanted to figure out what parts from the CI Test Ingest class I could take and put into generator and loader.


jkrdev:
I expect failure on the Travis build.

bfach10:
From the ticket description I'm interpreting that the first portion of this is to mock behavior of bulk ingesting files that are being continuously placed in a directory from MapReduce jobs (that outputs in AccumuloOutputFormat?)

With that assumption I would think the portion of the CI Test Ingest class that produces mutations for ingest and adds them to a BatchWriter won't help (mutations produced around line 160 of core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java). Is this the main reason why you mentioned you'll be deleting the ingest class?

jkrdev:
Right I still have to look more into it but I plan to strip away what I can from ContinuousIngest.java and put them into two new classes BulkGenerator.java and BulkLoader.java. I plan to have the Generator do all processes up to sending files to the Loader which will create the RFiles and load them into Accumulo. And like the CI test I want both of these to run until cancelled by the user. So both will be different applications (bl-local <applications>).

bfach10:
Ah ok. For the Generator this class in accumulo-examples may be of help: [BulkIngestExample.java](https://github.com/apache/accumulo-examples/blob/master/src/main/java/org/apache/accumulo/examples/mapreduce/bulk/BulkIngestExample.java). You'd probably have to remove the part where it has the shell run `importDirectory` so you can delegate this to the Loader (see below calls).

`FsShell fsShell = new FsShell(conf);`
` fsShell.run(new String[] {"-chmod", "-R", "777", opts.workDir});`
`connector.tableOperations().importDirectory(opts.getTableName(), opts.workDir + "/files", opts.workDir + "/failures", false);`

Maybe the generator could call a similar job in a `while(true)` loop like that in the CI test to get the "run until user cancel" functionality you mentioned?

jkrdev:
Thanks for the suggestion I had forgotten all about example will help a lot moving forward.

keith-turner:
@jkrdev  I posted a possible high level design on the JIRA issue.  Let me know if you have any questions.

I don't think any classes need to be copied.  I think just a few classes need to be added to core/src/main/java/org/apache/accumulo/testing/core/continuous that can support bulk loading instead of  loading via batch writer.

jkrdev:
Thanks I saw your comment, I talked to @ctubbsii about this earlier on and that is about the same design plan we came up with. 

Now on the point of no classes being copied, since its not continuous ingest I still think having them in separate packages is a better idea. I simply have a .../testing/core/bulk where all the files for bulk loading are. But I am still looking into this as a whole. Figuring out how to do each necessary step.

keith-turner:
> since its not continuous ingest 

Depends on how you think of it.  Could be continuous ingest via bulk load or continuous ingest via batch writer..

jkrdev:
ah okay, I think I understand. going to leave a comment on the jira ticket.

jkrdev:
Does anyone know what Maven command travis runs for this? I ran `mvn clean verify -DskipITs` and it passed locally but since it's failing here maybe that isn't the command being run?

ctubbsii:
@jkrdev Saw you asked, but can't find it on GitHub now; maybe you already figured it out and deleted your question, but in case you're still wondering, the command Travis CI uses to build the Maven project can be found in the `.travis.yml` file. Currently, it is `mvn clean verify`

jkrdev:
My bad I deleted it because Ben told me where to find it. I also realized the Travis build only failed because it couldn't connect not something on my end. Though should be noted that it will build cleanly but it not actually done. Would love a review though if possible. At the moment I am working on the generator class need to figure out how to have multiple generators that create the data (numbers) and the link list and put then in a directory, the loader class can then use that directory to make rfiles and the rest of the steps.

milleruntime:
This was done for ContinuousIngest in 8f6c24cbc1153515b76703a0395eadff80bc620f

PR #12, opened 2018-04-27
Updates to Undefined Analyzer made for apache/accumulo#441

keith-turner:
The code changed in this PR used to directly parse the master logs.  I changed the java code to not parse the master log, but rely on a shell command to get the needed info form the master log.


keith-turner:
If white space changes are ignored, then these changes are much smaller.

PR #15, opened 2018-06-27
Added random pausing to continuous ingest

mikewalch:


core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+  private static long lastPause;
+  private static long pauseWaitMs;
 
   private static ColumnVisibility getVisibility(Random rand) {
     return visibilities.get(rand.nextInt(visibilities.size()));
   }
 
+  private static boolean pauseEnabled(Properties props) {
+    String value = props.getProperty(TestProps.CI_INGEST_PAUSE_ENABLED);
+    return value != null && value.equals("true");
```
ctubbsii: Could use `return Boolean.parseBoolean(value);` here.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+  }
+
+  private static int getPauseWaitMs(Properties props, Random rand) {
+    int waitMin = Integer.parseInt(props.getProperty(TestProps.CI_INGEST_PAUSE_WAIT_MIN));
+    int waitMax = Integer.parseInt(props.getProperty(TestProps.CI_INGEST_PAUSE_WAIT_MAX));
```
ctubbsii: Should probably check args to make sure `waitMax >= waitMin` and that both are `>=0`.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+    return (rand.nextInt(waitMax - waitMin) + waitMin) * 1000;
+  }
+
+  private static int getPauseDurationMs(Properties props, Random rand) {
+    int durationMin = Integer.parseInt(props.getProperty(TestProps.CI_INGEST_PAUSE_DURATION_MIN));
+    int durationMax = Integer.parseInt(props.getProperty(TestProps.CI_INGEST_PAUSE_DURATION_MAX));
```
ctubbsii: Should check that `max>=min` and both `>=0`

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+    return (rand.nextInt(durationMax - durationMin) + durationMin) * 1000;
```
ctubbsii: `rand.nextInt(0)` throws an exception. Should check that max is strictly greater than min before using the difference in the `rand.nextInt()`.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+  }
+
+  private static void pauseCheck(Properties props, Random rand) throws InterruptedException {
+    if (pauseEnabled(props)) {
+      if (System.currentTimeMillis() > (lastPause + pauseWaitMs)) {
+        long pauseDurationMs = getPauseDurationMs(props, rand);
+        System.out.println("Pausing ingest for " + pauseDurationMs + " ms - " + System.currentTimeMillis());
+        Thread.sleep(pauseDurationMs);
+        lastPause = System.currentTimeMillis();
```
ctubbsii: Should use nanoTime for time durations.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -45,11 +45,43 @@
...
+        pauseWaitMs = getPauseWaitMs(props, rand);
+        System.out.println("Ingest restarted. Will pause again in " + pauseWaitMs + " ms at " + (lastPause + pauseWaitMs));
```
ctubbsii: Is STDOUT consistent with the rest of CI output, or should we use a logger?

mikewalch: STDOUT was consistent but I think a logger is better so I switched everything.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
mikewalch: Using nanoTime now.

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
mikewalch: Fixed in 84fcd3a630

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
mikewalch: Fixed in 84fcd3a630

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
mikewalch: Fixed in 84fcd3a630

core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
mikewalch: I now return min if they are equal.

PR #21, opened 2018-07-20
Created performance test framework

keith-turner:


keith-turner:
This change depends on apache/fluo-uno#185

keith-turner:
Below is an example output from running one of the test.

```json
{
  "testClass": "org.apache.accumulo.testing.core.performance.tests.ScanExecutorPT",
  "accumuloVersion": "2.0.0-SNAPSHOT",
  "startTime": "2018-07-20T03:20:07.178Z",
  "finishTime": "2018-07-20T03:27:41.840Z",
  "id": "sexec",
  "description": "Scan Executor Test.  Test running lots of short scans while long scans are running in the background.  Each short scan reads a random row and family. A scan prioritizer that favors short scans is configured.  If the scan prioritizer is not working properly, then the short scans will be orders of magnitude slower.",
  "results": [
    {
      "id": "write",
      "data": 849617.6720475786,
      "description": "Data write rate entries/sec ",
      "purpose": "COMPARISON"
    },
    {
      "id": "compact",
      "data": 283527.0768358378,
      "description": "Compact rate entries/sec ",
      "purpose": "COMPARISON"
    },
    {
      "id": "short_times1",
      "stats": {
        "min": 1,
        "max": 244,
        "sum": 214100,
        "average": 4.282,
        "count": 50000
      },
      "description": "Times in ms for each short scan.  First run.",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "short_times2",
      "stats": {
        "min": 1,
        "max": 127,
        "sum": 2012445,
        "average": 4.02489,
        "count": 500000
      },
      "description": "Times in ms for each short scan. Second run.",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "short",
      "data": 4.02489,
      "description": "Average times in ms for short scans from 2nd run.",
      "purpose": "COMPARISON"
    },
    {
      "id": "long_counts",
      "stats": {
        "min": 1001,
        "max": 7001,
        "sum": 201050,
        "average": 4021.0,
        "count": 50
      },
      "description": "Entries read by each long scan threads",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "long",
      "data": 450.7291701695979,
      "description": "Combined rate in entries/second of all long scans",
      "purpose": "COMPARISON"
    }
  ],
  "parameters": [
    {
      "id": "short_threads",
      "data": "5",
      "description": "Threads used to run short scans."
    },
    {
      "id": "long_threads",
      "data": "50",
      "description": "Threads running long scans.  Each thread repeatedly scans entire table for duration of test."
    },
    {
      "id": "rows",
      "data": "10000",
      "description": "Rows in test table"
    },
    {
      "id": "familes",
      "data": "10",
      "description": "Families per row in test table"
    },
    {
      "id": "qualifiers",
      "data": "10",
      "description": "Qualifiers per family in test table"
    },
    {
      "id": "server_scan_threads",
      "data": "2",
      "description": "Server side scan handler threads"
    },
    {
      "id": "prioritizer",
      "data": "org.apache.accumulo.core.spi.scan.IdleRatioScanPrioritizer",
      "description": "Server side scan prioritizer"
    }
  ]
}

```

bin/performance-test
```diff
@@ -0,0 +1,97 @@
...
+case "$1" in
+  run)
+    if [ -z "$2" ]; then
+      echo "ERROR: <output dir> needs to be set"
+      print_usage
+      exit 1
+    fi
+    mkdir -p "$2"
+    start_cluster
+    java -cp "$CP" -Dlog4j.configuration="file:$log4j_config" ${perf_pkg}.ListTests | while read test_class; do
```
ctubbsii: Using the following instead of the `| while read` syntax will fix the issue you're seeing with the file descriptor output errors.
```bash
readarray -t testlist < <(java -cp "$CP" -Dlog4j.configuration="file:$log4j_config" ${perf_pkg}.ListTests)
for t in "${testlist[@]}"; do
```

ctubbsii: Also, it's much better to `CLASSPATH=$CP; export CLASSPATH` than to use a potentially spammy long process line with `-cp "$CP"`

keith-turner: I was exporting classpath, but I moved this when I was not sure why uno was failing.   I would like to minimize what is exported by this script because some other env exports were borking uno (and figuring that out was time consuming, that is the reason this script does not use accumulo-testing-env.sh).  It would be nice to minimize the process line and not export, so I will try setting CLASSPATH just for the java processes.

core/src/main/java/org/apache/accumulo/testing/core/performance/Result.java
```diff
@@ -0,0 +1,30 @@
...
+public class Result {
+
+  public final String id;
+  public final Number data;
+  public final Stats stats;
+  public final String description;
+  public final Purpose purpose;
+
+  public enum Purpose {
+    INFORMATIONAL, COMPARISON
```
mikewalch: It might nice to add some comments describing the difference between informational and comparison.

PR #24, opened 2018-07-25
Add performance test for small scans

keith-turner:


keith-turner:
This PR copied and modified the following code.

https://github.com/keith-turner/smalls

PR #34, opened 2018-09-21
Replace Connector with AccumuloClient

milleruntime:


core/src/main/java/org/apache/accumulo/testing/core/scalability/ScaleTest.java
```diff
@@ -48,8 +51,7 @@ public void init(Properties scaleProps, Properties testProps, int numTabletServe
     String password = this.scaleProps.getProperty("PASSWORD");
     System.out.println(password);
 
-    conn = new ZooKeeperInstance(ClientConfiguration.create().withInstance(instanceName).withZkHosts(zookeepers)).getConnector(user,
-        new PasswordToken(password));
+    Accumulo.newClient().forInstance(instanceName, zookeepers).usingToken(user, new PasswordToken(password)).build();
```
mikewalch: it looks like `conn` is not set here

core/src/main/java/org/apache/accumulo/testing/core/scalability/ScaleTest.java
```diff
@@ -79,7 +81,7 @@ protected void stopTimer(long numEntries, long numBytes) {
     return keys;
   }
 
-  public Connector getConnector() {
+  public AccumuloClient getConnector() {
```
mikewalch: this could be renamed to `getClient`

core/src/main/java/org/apache/accumulo/testing/core/performance/impl/PerfTestRunner.java
```diff
@@ -42,7 +43,7 @@ public static void main(String[] args) throws Exception {
 
     PerformanceTest perfTest = Class.forName(className).asSubclass(PerformanceTest.class).newInstance();
 
-    Connector conn = Connector.builder().usingProperties(clientProps).build();
+    Connector conn = Connector.from(Accumulo.newClient().usingProperties(clientProps).build());
```
mikewalch: Should this be switched over to `AccumuloClient`

PR #35, opened 2018-09-25
Updates due to Accumulo 2.0 changes

mikewalch:
- using accumulo-client.properties to create clients


core/src/main/java/org/apache/accumulo/testing/core/randomwalk/multitable/CopyTool.java
```diff
@@ -64,70 +56,21 @@ public int run(String[] args) throws Exception {
...
 
-    ClientConfiguration clientConf = ClientConfiguration.create().withInstance(args[3]).withZkHosts(args[4]);
-
+    ClientInfo info = Accumulo.newClient().usingProperties(args[0]).info();
     job.setInputFormatClass(AccumuloInputFormat.class);
-    AccumuloInputFormat.setInputTableName(job, args[2]);
+    AccumuloInputFormat.setClientInfo(job, info);
+    AccumuloInputFormat.setInputTableName(job, args[1]);
     AccumuloInputFormat.setScanAuthorizations(job, Authorizations.EMPTY);
-    AccumuloInputFormat.setZooKeeperInstance(job, clientConf);
```
keith-turner: Wow, what a difference!

core/src/main/java/org/apache/accumulo/testing/core/randomwalk/multitable/CopyTool.java
```diff
@@ -64,70 +56,21 @@ public int run(String[] args) throws Exception {
...
-        // to get a delegation token
-        if (!conn.securityOperations().hasSystemPermission(conn.whoami(), SystemPermission.OBTAIN_DELEGATION_TOKEN)) {
-          log.error(principal + " doesn't have the " + SystemPermission.OBTAIN_DELEGATION_TOKEN.name()
-              + " SystemPermission neccesary to obtain a delegation token. MapReduce tasks cannot automatically use the client's"
-              + " credentials on remote servers. Delegation tokens provide a means to run MapReduce without distributing the user's credentials.");
-          throw new IllegalStateException(conn.whoami() + " does not have permission to obtain a delegation token");
-        }
-
-        // Fetch a delegation token from Accumulo
-        token = conn.securityOperations().getDelegationToken(new DelegationTokenConfig());
```
keith-turner: So this isn't needed anymore?

mikewalch: The delegation token should be created outside of this code and set in the client properties that are passed into the MapReduce code.

keith-turner: > The delegation token should be created outside of this code and set in the client properties that are passed into the MapReduce code.

That is a more elegant way to use delegation tokens.  Then there is no need to write special code for each map reduce job.

PR #39, opened 2018-10-12
fixed peformance test

keith-turner:


core/src/main/java/org/apache/accumulo/testing/core/continuous/ContinuousIngest.java
```diff
@@ -129,7 +130,7 @@ public static void main(String[] args) throws Exception {
     }
 
     BatchWriter bw = client.createBatchWriter(tableName);
-    bw = Trace.wrapAll(bw, new CountSampler(1024));
+    bw = Trace.wrapAll(bw, TraceSamplers.countSampler(1024));
```
milleruntime: What changed here?  I guess I missed this update...

ctubbsii: That was to fix a spotbugs issue involving a subclass name masking the superclass name. I deleted our "convenience" classes which trivially subclassed htrace classes, and just turned them into convenience utility methods.

keith-turner: The test are using internal Accumulo code that changed.   Need to move the test away from using Accumulo code that is not public API.  I had to fix this inorder for things to compile even though it was completely unrelated to what I was doing.

PR #43, opened 2018-11-26
Major refactor to support running in Docker

mikewalch:
* Multiple test clients are now run in Docker instead of YARN
* Tests can still be run locally but accumulo-testing script
  has been removed and replaced by seperate scripts for
  cingest, rwalk, and agitator
* Project no longer contains multiple Maven modules
* Renamed package to remove 'core' from it
* Renamed accumulo-testing.env to env.sh
* Created Dockerfile


pom.xml
```diff
@@ -258,4 +151,122 @@
...
+      <version>${accumulo.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.accumulo</groupId>
+      <artifactId>accumulo-test</artifactId>
+      <version>${accumulo.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.hadoop</groupId>
```
keith-turner: For hadoop should only have the following two dependencies.  This is the new shaded jar with no deps, using it really cleans up the conflicts in the shaded jar.

```xml
    <dependency>
      <groupId>org.apache.hadoop</groupId>
      <artifactId>hadoop-client-api</artifactId>
      <version>${hadoop.version}</version>
    </dependency>
    <dependency>
      <groupId>org.apache.hadoop</groupId>
      <artifactId>hadoop-client-runtime</artifactId>
      <version>${hadoop.version}</version>
    </dependency>
```

pom.xml
```diff
@@ -27,99 +27,23 @@
...
-        <artifactId>hadoop-mapreduce-client-jobclient</artifactId>
-        <version>${hadoop.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>org.apache.hadoop</groupId>
-        <artifactId>hadoop-common</artifactId>
-        <version>${hadoop.version}</version>
-      </dependency>
       <dependency>
         <groupId>org.apache.hadoop</groupId>
```
keith-turner: May be able to remove hadoop stuff from dep mgmt when using new hadoop shaded jar.

pom.xml
```diff
@@ -258,4 +151,122 @@
...
+      <groupId>org.apache.accumulo</groupId>
+      <artifactId>accumulo-client-mapreduce</artifactId>
+      <version>${accumulo.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.accumulo</groupId>
+      <artifactId>accumulo-core</artifactId>
+      <version>${accumulo.version}</version>
+    </dependency>
+    <dependency>
```
keith-turner: It would be really nice to set the scope of this to `test` because mini brings a lot of Accumulo and Hadoop server side cruft into the shaded jar.  I tried doing this and ran into a problem with Replication test using Accumulo server side code.  This mini dep brings in that server side code.  For local testing I deleted the Replication test code and was able to build a shaded jar that had much less warnings.

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+ADD ./conf/accumulo-client.properties /opt/conf/
+ADD ./conf/accumulo-testing.properties /opt/conf/
+ADD ./conf/log4j.properties.example /opt/conf/
+ADD ./bin/cingest /opt/bin
+ADD ./bin/rwalk /opt/bin
+ADD ./bin/run /opt/bin
+ADD ./target/accumulo-testing-${TEST_JAR_VERSION}-shaded.jar /opt/
+
+ENTRYPOINT ["/opt/bin/run"]
```
keith-turner: Could give `run` script a more specific name like `docker-entry`.

bin/run
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+bin_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+at_home=$( cd "$( dirname "$bin_dir" )" && pwd )
+
+function print_usage() {
+  cat <<EOF
+
+Usage: accumulo-testing <script> (<argument>)
```
keith-turner: Since this script is only needed by docker could put it in `src/main/docker` and have the docker build file pull it from there.

PR #44, opened 2018-11-27
Build shaded jar without Hadoop for M/R

mikewalch:
* Hadoop jars get added to CLASSPATH when needed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,12 +18,9 @@ FROM centos:7
 RUN yum install -y java-1.8.0-openjdk-devel
 ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk
 
-ENV HADOOP_VERSION 3.1.1
-ENV TEST_JAR_VERSION 2.0.0-SNAPSHOT
-
-ENV HADOOP_API_JAR /opt/hadoop-client-api-${HADOOP_VERSION}.jar
-ENV HADOOP_RUNTIME_JAR /opt/hadoop-client-runtime-${HADOOP_VERSION}.jar
-ENV TEST_JAR_PATH /opt/accumulo-testing-${TEST_JAR_VERSION}-shaded.jar
+ENV HADOOP_API_JAR /opt/hadoop-client-api.jar
+ENV HADOOP_RUNTIME_JAR /opt/hadoop-client-runtime.jar
+ENV TEST_JAR_PATH /opt/accumulo-testing-shaded.jar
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -18,8 +18,11 @@ FROM centos:7
 RUN yum install -y java-1.8.0-openjdk-devel
 ENV JAVA_HOME /usr/lib/jvm/java-1.8.0-openjdk
 
-ARG TEST_JAR_VERSION
+ENV HADOOP_VERSION 3.1.1
```
keith-turner: Seems like this should be an argument

keith-turner: Or could have the maven dependency copy strip the versions.  Then dockerfile does not need to know versions.

keith-turner: cservice

mikewalch: Fixed now

Dockerfile
```diff
@@ -18,9 +18,9 @@ FROM centos:7
...
 
-ARG TEST_JAR_VERSION
-ENV TEST_JAR_VERSION 2.0.0-SNAPSHOT
-ENV TEST_JAR_PATH /opt/accumulo-testing-${TEST_JAR_VERSION}-shaded.jar
+ENV HADOOP_API_JAR /opt/hadoop-client-api.jar
+ENV HADOOP_RUNTIME_JAR /opt/hadoop-client-runtime.jar
+ENV TEST_JAR_PATH /opt/accumulo-testing-shaded.jar
 ENV ACCUMULO_CLIENT_PROPS /opt/conf/accumulo-client.properties
 ENV TEST_PROPS /opt/conf/accumulo-testing.properties
 ENV TEST_LOG4J /opt/conf/log4j.properties.example
```
keith-turner: Could copy `log4j.properties.example` and `log4j.properties` into docker file and have script fallback to example if regular file does not exists.   I think the docker `COPY` command can take multiple files.

Dockerfile
```diff
@@ -35,7 +35,9 @@ ADD ./conf/log4j.properties.example /opt/conf/
 ADD ./bin/cingest /opt/bin
 ADD ./bin/rwalk /opt/bin
 ADD ./src/main/docker/docker-entry /opt/bin
-ADD ./target/accumulo-testing-${TEST_JAR_VERSION}-shaded.jar /opt/
+ADD ./target/accumulo-testing-shaded.jar /opt/
```
keith-turner: https://stackoverflow.com/a/24958548/7298689

mikewalch: Done. See ffc78ca9c0c

PR #46, opened 2018-11-29
Closed clients that weren't being closed

mikewalch:
* Stopped using changeUser
* Closed clients using try-with-resources blocks


mikewalch:
Hide whitespace changes when you review this.

PR #49, opened 2019-01-02
Updates for starting MapReduce jobs from randomwalk

mikewalch:
* Set additional MapReduce configuration and Hadoop username
* These settings work if running randomwalk on user machine
  or in Docker

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/TestEnv.java
+++ b/src/main/java/org/apache/accumulo/testing/TestEnv.java
@@ -97,6 +97,13 @@ public class TestEnv implements AutoCloseable {
       hadoopConfig.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
       hadoopConfig.set("mapreduce.framework.name", "yarn");
       hadoopConfig.set("yarn.resourcemanager.hostname", getYarnResourceManager());
+      String hadoopHome = System.getenv("HADOOP_HOME");
+      if (hadoopHome == null) {
+        throw new IllegalArgumentException("HADOOP_HOME must be set in env");
+      }
+      hadoopConfig.set("yarn.app.mapreduce.am.env", "HADOOP_MAPRED_HOME=" + hadoopHome);
+      hadoopConfig.set("mapreduce.map.env", "HADOOP_MAPRED_HOME=" + hadoopHome);
+      hadoopConfig.set("mapreduce.reduce.env", "HADOOP_MAPRED_HOME=" + hadoopHome);
     }
     return hadoopConfig;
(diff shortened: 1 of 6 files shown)


src/main/java/org/apache/accumulo/testing/TestEnv.java
```diff
@@ -97,6 +97,13 @@ public Configuration getHadoopConfiguration() {
       hadoopConfig.set("fs.file.impl", org.apache.hadoop.fs.LocalFileSystem.class.getName());
```
keith-turner: Unrelated to your change, but this seems really odd.

src/main/java/org/apache/accumulo/testing/TestEnv.java
```diff
@@ -97,6 +97,13 @@ public Configuration getHadoopConfiguration() {
...
       hadoopConfig.set("mapreduce.framework.name", "yarn");
       hadoopConfig.set("yarn.resourcemanager.hostname", getYarnResourceManager());
+      String hadoopHome = System.getenv("HADOOP_HOME");
+      if (hadoopHome == null) {
+        throw new IllegalArgumentException("HADOOP_HOME must be set in env");
+      }
+      hadoopConfig.set("yarn.app.mapreduce.am.env", "HADOOP_MAPRED_HOME=" + hadoopHome);
```
keith-turner: I have been searching for a long time to no avail trying to understand what the expectations for these settings are and what best practices are for Hadoop.  It seems odd that a client submitting a job has to know where the map reduce jars are on the cluster.

I am curious if `HADOOP_MAPRED_HOME` is set in `hadoop-env.sh` on the cluster if the client does not have to set it.  I have not been able to find any definitive docs on this.  If I get a chance I may try experimenting with this.

src/main/java/org/apache/accumulo/testing/TestEnv.java
ctubbsii: This is the implementation to handle HDFS filesystems that are `file://` URI-based instead of `hdfs://`. It's only ever used if the configured DFS filesystem is local.

keith-turner: Thanks @ctubbsii .  I knew that but had forgotten.

src/main/java/org/apache/accumulo/testing/TestEnv.java
mikewalch: `HADOOP_MAPRED_HOME` is being set in `hadoop-env.sh` by uno.  It did not seem to make a difference when I ran tests.

PR #51, opened 2019-01-25
Improvements made for 2.0.0-alpha-2 CI testing

keith-turner:
Added ability to set test options on command line. This makes running different
types of ingest from the same docker image easy.  Without this change a new
docker image would need to be created for each different ingest type.

Documented commands for running a long running ingest test.

Made the ingest flush size configurable.  This made it possible to write really
small amounts of data and then pause.  This changes makes testing
for the problem in apache/accumulo#854 possible.

Added docs for how to copy docker image to other nodes if you don't have
a repository handy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,13 +66,18 @@ run in Docker:
 3. Multiple containers can also be run (if you have [Docker Swarm] enabled):
 
    ```bash
+   # the following can be used to get the image on all nodes if you do not have a registry.
+   for HOST in node1 node2 node3; do
+     docker save accumulo-testing | ssh -C $HOST docker load &
+   done
+
    docker service create --network="host" --replicas 2 --name ci accumulo-testing cingest ingest
    ```
 
 ## Random walk test
 
 The random walk test generates client behavior on an Apache Accumulo instance by randomly walking a
-graph of client operations. 
+graph of client operations.
 
 Before running random walk, review the `test.common.*` properties in `accumulo-testing.properties`
 file. A test module must also be specified. See the [modules] directory for a list of available ones.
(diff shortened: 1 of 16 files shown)


keith-turner:
Best to avoid whitespace when looking at this diff.

README.md
```diff
@@ -66,13 +66,18 @@ run in Docker:
 3. Multiple containers can also be run (if you have [Docker Swarm] enabled):
 
    ```bash
+   # the following can be used to get the image on all nodes if you do not have a registry.
+   for HOST in node1 node2 node3; do
+     docker save accumulo-testing | ssh -C $HOST docker load &
```
mikewalch: very cool

bin/agitator
```diff
@@ -29,10 +29,10 @@ Possible commands:
 EOF
 }
 
-if [ -f "$at_home/conf/accumulo-testing-env.sh" ]; then
-  . "$at_home"/conf/accumulo-testing-env.sh
```
mikewalch: nice catch

pom.xml
```diff
@@ -33,7 +33,7 @@
   <description>Testing tools for Apache Accumulo</description>
 
   <properties>
-    <accumulo.version>2.0.0-SNAPSHOT</accumulo.version>
+    <accumulo.version>2.0.0-alpha-2</accumulo.version>
```
mikewalch: Maybe this should be left at 2.0.0-SNAPSHOT

README.md
```diff
@@ -101,12 +106,12 @@ First, run the command below to create an Accumulo table for the continuous inge
 table is set by the property `test.ci.common.accumulo.table` (its value defaults to `ci`) in the file
 `accumulo-testing.properties`:
 
-          ./bin/cingest createtable
+          ./bin/cingest createtable {-o <prop>=<value>}
```
mikewalch: could mention that only `test.*` props can be overriden

PR #52, opened 2019-01-29
Update checkstyle to use import-control for API checking

milleruntime:
This should be easier to maintain than the regex (at least for 2.0) and the output is a lot more helpful:
<pre>
[ERROR] src/main/java/org/apache/accumulo/testing/stress/ScanOpts.java:[19,1] (imports) ImportControl: Disallowed import - org.apache.accumulo.core.cli.ClientOnDefaultTable.
[ERROR] src/main/java/org/apache/accumulo/testing/stress/WriteOptions.java:[19,1] (imports) ImportControl: Disallowed import - org.apache.accumulo.core.cli.ClientOnDefaultTable.
[ERROR] src/main/java/org/apache/accumulo/testing/stress/Write.java:[19,1] (imports) ImportControl: Disallowed import - org.apache.accumulo.core.cli.BatchWriterOpts.

</pre>
VS the old output
<pre>
[ERROR] src/main/java/org/apache/accumulo/testing/stress/ScanOpts.java:[19] (regexp) RegexpSinglelineJava: Accumulo non-public classes imported
[ERROR] src/main/java/org/apache/accumulo/testing/stress/WriteOptions.java:[19] (regexp) RegexpSinglelineJava: Accumulo non-public classes imported
[ERROR] src/main/java/org/apache/accumulo/testing/stress/Write.java:[19] (regexp) RegexpSinglelineJava: Accumulo non-public classes imported
</pre>

There are a lot of crazy things you can do with the [import-control checkstyle rule](http://checkstyle.sourceforge.net/config_imports.html#ImportControl) but this was the best I could come up with... 

I also upgraded the checkstyle plugin to 3.0.0 and made it use the latest checkstyle dependency 8.17.

### DIFF
--- a/contrib/import-control.xml
+++ b/contrib/import-control.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+      http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<!DOCTYPE import-control PUBLIC
+    "-//Checkstyle//DTD ImportControl Configuration 1.4//EN"
+    "https://checkstyle.org/dtds/import_control_1_4.dtd">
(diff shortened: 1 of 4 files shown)


contrib/import-control.xml
```diff
@@ -0,0 +1,31 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<!DOCTYPE import-control PUBLIC
+    "-//Checkstyle//DTD ImportControl Configuration 1.4//EN"
+    "https://checkstyle.org/dtds/import_control_1_4.dtd">
+<import-control pkg="org.apache.accumulo.testing" strategyOnMismatch="allowed">
```
keith-turner: This config is really nice, much easier to understand than the regex.  We could adopt a strategy of holding the line where we add exceptions for every non-API currently used and always run this check on build.  Then we can remove usage of non-public API going forward and prevent new usages of non-public API.

PR #55, opened 2019-02-07
Update japi readme and master config

milleruntime:


milleruntime:
> The accumulo-client-mapreduce jar is wrong.

Where are you seeing this?

ctubbsii:
> > The accumulo-client-mapreduce jar is wrong.
> 
> Where are you seeing this?

https://github.com/apache/accumulo-testing/blob/f9632f9ef05f25479305169f938bf57f4389d87e/test/compat/japi-compliance/japi-accumulo-master.xml#L30

milleruntime:
Thank you!  I had updated that file for my local paths but didn't want to push it.  Then your comment I was like "I know I have seen that and fixed it!"

PR #56, opened 2019-02-08
Added perf test for yielding+scan dispatching

keith-turner:
Test running lots of short scans while many filters that return little
data are running in the background.  If nothing is done these filters
will prevent the short scans from running.  This test configures
Accumulo so that the short scans should get a chance to run.  It does
three things to facilitate the short scans : the filters yield, there
are two scan executors, and a scan dispatcher sends long running scans
to the second executor. If yielding and dispatching are working
correctly then the short scans should have very short response times.
This happens because the filters should end up in a separate thread pool
than the short scan.

Also fixed some bugs with the performance test framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,11 @@ To run performance test a `cluster-control.sh` script is needed to assist with s
 wiping, and confguring an Accumulo instance.  This script should define the following functions.
 
 ```bash
+
+function get_hadoop_client {
+  # TODO return hadoop client libs in a form suitable for appending to a classpath
+}
+
 function get_version {
   case $1 in
     ACCUMULO)
--- a/bin/performance
+++ b/bin/performance
@@ -28,12 +28,13 @@ Possible commands:
   run <output dir> [filter]        Runs performance tests.
   compare <result 1> <result 2>    Compares results of two test.
   csv {files}                      Converts results to CSV
+  list                             List the performance test
 EOF
 }
 
 
 function build_shade_jar() {
-  at_shaded_jar="$at_home/core/target/accumulo-testing-$at_version-shaded.jar"
+  at_shaded_jar="$at_home/target/accumulo-testing-shaded.jar"
   if [ ! -f "$at_shaded_jar" ]; then
     echo "Building $at_shaded_jar"
     cd "$at_home" || exit 1
@@ -58,7 +59,8 @@ fi
 
 . $at_home/conf/cluster-control.sh
 build_shade_jar
-CP="$at_home/core/target/accumulo-testing-$at_version-shaded.jar"
+# TODO be more specific with hadoop jars
+CP="$at_home/target/accumulo-testing-shaded.jar:$HADOOP_HOME/share/hadoop/client/*"
 perf_pkg="org.apache.accumulo.testing.performance.impl"
 case "$1" in
   run)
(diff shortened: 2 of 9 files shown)


keith-turner:
Below is the output of running this test on a single node Accumulo instance.  Saw 4.6ms avg response times from the 5 threads doing 100K random small lookups. While the 100K lookups were runnning 50 threads were running filters that randomly returned 1/1,000,000 of the data.  Normally these filters would kill the performance of the small lookups.

```json
{
  "testClass": "org.apache.accumulo.testing.performance.tests.YieldingScanExecutorPT",
  "accumuloVersion": "2.0.0-alpha-2",
  "startTime": "2019-02-08T19:10:41.752Z",
  "finishTime": "2019-02-08T19:15:27.644Z",
  "id": "yfexec",
  "description": "Scan Executor Test.  Test running lots of short scans while many filters that return little data are running in the background.  If nothing is done these filters will prevent the short scans from running.  This test configures Accumulo so that the short scans should get a chance to run.  It does three things to facilitate the short scans : the filters yield, there a two scan executors, and a scan dispatcher sends long running scans to the second executor. If yielding and dispatching are working correctly then the short scans should have very short response times.  This happens because the filters should end up in a separate thread pool than the short scan.",
  "results": [
    {
      "id": "write",
      "data": 1532801.9619865112,
      "description": "Data write rate entries/sec ",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "compact",
      "data": 410273.24197915813,
      "description": "Compact rate entries/sec ",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "short_times1",
      "stats": {
        "min": 1,
        "max": 42770,
        "sum": 447442,
        "average": 8.94884,
        "count": 50000
      },
      "description": "Times in ms for each short scan.  First run.",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "short_times2",
      "stats": {
        "min": 1,
        "max": 337,
        "sum": 456498,
        "average": 4.56498,
        "count": 100000
      },
      "description": "Times in ms for each short scan. Second run.",
      "purpose": "INFORMATIONAL"
    },
    {
      "id": "short",
      "data": 4.56498,
      "description": "Average times in ms for short scans from 2nd run.",
      "purpose": "COMPARISON"
    },
    {
      "id": "long_counts",
      "stats": {
        "min": 2,
        "max": 23,
        "sum": 427,
        "average": 8.54,
        "count": 50
      },
      "description": "Entries read by each of the filter threads",
      "purpose": "INFORMATIONAL"
    },
    "id": "long",
      "data": 2.3583342538385064,
      "description": "Combined rate in entries/second of all long scans.  This should be low but non-zero.",
      "purpose": "INFORMATIONAL"
    }
  ],
  "parameters": [
    {
      "id": "short_threads",
      "data": "5",
      "description": "Threads used to run short scans."
    },
    {
      "id": "long_threads",
      "data": "50",
      "description": "Threads running long fileter scans.  Each thread repeatedly scans entire table for duration of test randomly returning a few of the keys."
    },
    {
      "id": "rows",
      "data": "100000",
      "description": "Rows in test table"
    },
    {
      "id": "familes",
      "data": "10",
      "description": "Families per row in test table"
    },
    {
      "id": "qualifiers",
      "data": "10",
      "description": "Qualifiers per family in test table"
    },
    {
      "id": "server_scan_threads",
      "data": "2",
      "description": "Server side scan handler threads that each executor has.  There are 2 executors."
    },
    {
      "id": "filter_probability",
      "data": "0.000001",
      "description": "The chance that one of the long filter scans will return any key it sees."
    },
    {
      "id": "filter_yield_time",
      "data": "1000",
      "description": "The time in ms after which one of the long filter scans will yield."
    },
    {
      "id": "quick_scan_time",
      "data": "500",
      "description": "The threshold time in ms for deciding what is a quick vs long scan.  Times less than this are sent to one executor and longer times are sent to another."
    }
  ]
}
```

PR #60, opened 2019-02-26
Factor RFile and CryptoServiceFactory from Import-Control

Manno15:
Factored out RFile (small change) and CryptoServiceFactory. I took the needed functionality from the CryptoSerivceFactory class and created a small function inside testIngest where it was used.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/ingest/TestIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/ingest/TestIngest.java
@@ -35,8 +35,10 @@ import org.apache.accumulo.core.client.TableExistsException;
 import org.apache.accumulo.core.client.TableNotFoundException;
 import org.apache.accumulo.core.client.security.SecurityErrorCode;
 import org.apache.accumulo.core.clientImpl.TabletServerBatchWriter;
+import org.apache.accumulo.core.conf.AccumuloConfiguration;
 import org.apache.accumulo.core.conf.DefaultConfiguration;
-import org.apache.accumulo.core.crypto.CryptoServiceFactory;
+import org.apache.accumulo.core.conf.Property;
+import org.apache.accumulo.core.spi.crypto.CryptoService;
 import org.apache.accumulo.core.data.ConstraintViolationSummary;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Mutation;
@@ -44,7 +46,6 @@ import org.apache.accumulo.core.data.TabletId;
 import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.file.FileOperations;
 import org.apache.accumulo.core.file.FileSKVWriter;
-import org.apache.accumulo.core.file.rfile.RFile;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.accumulo.core.security.ColumnVisibility;
 import org.apache.accumulo.core.trace.DistributedTrace;
@@ -226,11 +227,8 @@ public class TestIngest {
     FileSKVWriter writer = null;
 
     if (opts.outputFile != null) {
-      writer = FileOperations
-          .getInstance()
-          .newWriterBuilder()
-          .forFile(opts.outputFile + "." + RFile.EXTENSION, fs, conf,
-              CryptoServiceFactory.newDefaultInstance())
+      writer = FileOperations.getInstance().newWriterBuilder()
+          .forFile(opts.outputFile + ".rf", fs, conf, CryptoServiceInstance())
           .withTableConfiguration(DefaultConfiguration.getInstance()).build();
       writer.startDefaultLocalityGroup();
     } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/accumulo/testing/ingest/TestIngest.java
```diff
@@ -226,11 +227,8 @@ public static void ingest(AccumuloClient client, FileSystem fs, Opts opts,
     FileSKVWriter writer = null;
 
     if (opts.outputFile != null) {
-      writer = FileOperations
-          .getInstance()
-          .newWriterBuilder()
-          .forFile(opts.outputFile + "." + RFile.EXTENSION, fs, conf,
-              CryptoServiceFactory.newDefaultInstance())
+      writer = FileOperations.getInstance().newWriterBuilder()
```
keith-turner: You could take this a little further.  There is a public API for creating Rfiles that could be used instead of FileOperations.  [RFile.newWriter()](https://static.javadoc.io/org.apache.accumulo/accumulo-core/2.0.0-alpha-2/org/apache/accumulo/core/client/rfile/RFile.html#newWriter--) is the entry point in the API for build an [RFileWriter](https://static.javadoc.io/org.apache.accumulo/accumulo-core/2.0.0-alpha-2/org/apache/accumulo/core/client/rfile/RFileWriter.html).  If this API is used then I think this class can avoid using FileSKVWriter, FileOperations, CryptoService, and possibly DefaultConfiguration.

Manno15: Yeah, I will take a look into that. Thanks for the idea.

src/main/java/org/apache/accumulo/testing/ingest/TestIngest.java
```diff
@@ -226,11 +227,8 @@ public static void ingest(AccumuloClient client, FileSystem fs, Opts opts,
...
+          .forFile(opts.outputFile + ".rf", fs, conf, CryptoServiceInstance())
```
milleruntime: These changes are better but I think this could be done using just the RFile user API.  Checkout [RFile](https://accumulo.apache.org/1.9/apidocs/org/apache/accumulo/core/client/rfile/RFile.html#newWriter--) and [RFileWriter](https://accumulo.apache.org/1.9/apidocs/org/apache/accumulo/core/client/rfile/RFileWriter.html).  This would eliminate the need to provide a CryptoService.

milleruntime: Sorry Keith beat me to the punch!

src/main/java/org/apache/accumulo/testing/ingest/TestIngest.java
Manno15: ` RFileWriter writer = null;
    if (opts.outputFile != null) {
      writer = RFile.newWriter().to(opts.outputFile + ".rf").withFileSystem(fs).build();`
Would something along the lines of this work or am I missing something?

ctubbsii: For what it's worth, RFileWriter is AutoCloseable, and could be used inside a try-with-resources block. Should at least ensure it's closed in a finally block, if not null.

Manno15: `try(RfileWriter writer = RFile.newWriter().to(opts.outputFile + ".rf").withFileSystem(fs).build()){`
This is similar to the example given. Just wasn't sure if it was necessary. How does that look?

Manno15: Sorry on the formatting of the code, I saw you edited the last one

ctubbsii: > Sorry on the formatting of the code, I saw you edited the last one

No apologies necessary. I did that partly for readability, and partly as an demo/example for how to do it in future. It's not necessary to format things that way. It's just helpful once you learn to use all these little hidden gems in GitHub-flavored markdown. :smiley_cat:

Manno15: @ctubbsii What did you have in mind with the try-with block? Currently how the code is formatted, it would take some shifting of the rest of the function code to get everything needed inside the block. For example, the startTime and generateRow is used by both the RFile writer and batch writer. Also, there is a check to make sure to close the writer at the end of the function but I am not sure if that suffices or not.

ctubbsii: It was more of a passing thought... a suggestion to check. If the code is structured relatively linearly, the try-with-resources construct can be a clean and elegant way to keep the lifecycle of AutoCloseable resources, such as the RFile writer, well-managed. I didn't dive deep enough to look at the existing code structure, though.

Manno15: Yeah, I understand. It was a good suggestion and I will definitely keep it in mind for future use.

keith-turner: @Manno15  that looks good.

mjwall:
Thanks @Manno15 , feel free to add your self to the contributors with PR in accumulo-website

keith-turner:
Can use the following link to add yourself as a contributor if interested.

https://github.com/apache/accumulo-website/edit/master/pages/people.md

PR #61, opened 2019-02-27
Clean up CLI parsing and stop depending on Accumulo internals

mikewalch:
* Remove dependencies on Accumulo internal CLI parsing code
  by addeing ClientOpts and Help
* Remove trace code from TestIngest

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
+++ b/src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
@@ -1,0 +1,213 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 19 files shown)


src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
```diff
@@ -0,0 +1,213 @@
...
+    return clientConfigFile;
+  }
+
+  public Properties getClientProperties() {
+    if (cachedProps == null) {
+      cachedProps = new Properties();
+      if (getClientConfigFile() != null) {
+        cachedProps = toProperties(Paths.get(getClientConfigFile()));
+      }
+      if (saslEnabled) {
```
keith-turner: this could be a follow on PR, but looking at this made me think of it.  I think it would be nice to remove any command line options that can be set in the client props.  Also could support a command line option to set a client prop.  I recently did this in #51 for continuous ingest code.  Being able to set client props on the command line is great for docker.

mikewalch: Nice idea. I look at doing this in another PR.

PR #64, opened 2019-03-08
Made CI bulk import configurable and scriptable

keith-turner:


docs/bulk-test.md
```diff
@@ -0,0 +1,35 @@
+# Running a bulk ingest test
+
+Continous ingest supports bulk ingest in addition to live ingest. A map reduce
+job that generates rfiles using the tables splits can be run.  This can be run
+in a loop like the following to continually bulk import data.
```
milleruntime: Could mention that it works with "cingest createtable" as well.

src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
```diff
@@ -103,157 +102,13 @@ public int run(String[] args) throws Exception {
...
         boolean success = job.isSuccessful();
 
-        // bulk import completed files
-        if (success) {
-          log.info("Sort and create job successful. Bulk importing {} to {}", RFILE_DIR, tableName);
-          client.tableOperations().importDirectory(RFILE_DIR).to(tableName).load();
-        } else {
-          log.error("Job failed, not calling bulk import");
-        }
         return success ? 0 : 1;
```
milleruntime: Looks like you don't need `success` anymore

keith-turner: I think its still useful for the process return code.

PR #65, opened 2019-03-14
Update ClientOpts

mikewalch:
* Reduce the number of options
* Create clients using Accumulo.newClient()
* Remove use MapReduce opts from Accumulo

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
+++ b/src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
@@ -24,10 +24,13 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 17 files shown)


src/main/java/org/apache/accumulo/testing/cli/ClientOpts.java
```diff
@@ -169,34 +165,20 @@ public String getClientConfigFile() {
...
-        cachedProps.setProperty(ClientProperty.INSTANCE_NAME.getKey(), instance);
-      }
       if (securePassword != null) {
         ClientProperty.setPassword(cachedProps, securePassword.toString());
-      } else if (password != null) {
-        ClientProperty.setPassword(cachedProps, password.toString());
-      } else if (keytabPath != null) {
-        ClientProperty.setKerberosKeytab(cachedProps, keytabPath);
       }
+      getOverrides().forEach((k, v) -> cachedProps.put(k, v));
```
ctubbsii: I wonder if `cachedProps::put` works instead of explicitly putting in the params.

mikewalch: Yes it works.

PR #68, opened 2019-03-20
Refactored RowHash and TeraSortIngest

mikewalch:


bin/mapred
```diff
@@ -0,0 +1,67 @@
...
+Usage: mapred <application> {-o test.<prop>=<value>}
+
+Available applications:
+
+    terasort    Run Terasort
+    rowhash     Run RowHash
+EOF
+}
+
+if [ -f "$at_home/conf/env.sh" ]; then
```
ctubbsii: Double-brackets keeps bash from executing `/usr/bin/[`, an external command.

In addition to that, may want to run shellcheck on the whole script, just to catch any bash "gotchas".

mikewalch: Created #69 to do this for all scripts

PR #74, opened 2019-05-20
Add Durability Performance Test to Fix Durability IT

Manno15:
This is a performance test to replace the test currently in Durability IT that was causing some issues. 
Link to that issue below. 
https://github.com/apache/accumulo/issues/1159

The test writeSpeed is better suited to a performance test than being in an IT that could cause failures depending on what hardware ran the test.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/DurabilityWriteSpeedPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/DurabilityWriteSpeedPT.java
@@ -1,0 +1,110 @@
+package org.apache.accumulo.testing.performance.tests;
+
+import java.util.Arrays;
+
+import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.client.BatchWriter;
+import org.apache.accumulo.core.client.admin.TableOperations;
+import org.apache.accumulo.core.conf.Property;
+import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.testing.performance.Environment;
+import org.apache.accumulo.testing.performance.PerformanceTest;
+import org.apache.accumulo.testing.performance.Report;
+import org.apache.accumulo.testing.performance.SystemConfiguration;
+
+public class DurabilityWriteSpeedPT implements PerformanceTest {
+  static final long N = 100000;
+
+  @Override
+  public SystemConfiguration getSystemConfig() {
+    return new SystemConfiguration();
+  }
+
+  private String[] init(AccumuloClient c) throws Exception {
+    String[] tableNames = getUniqueNames(4);
+    TableOperations tableOps = c.tableOperations();
+    createTable(c, tableNames[0]);
+    createTable(c, tableNames[1]);
+    createTable(c, tableNames[2]);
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/performance/tests/DurabilityWriteSpeedPT.java
```diff
@@ -0,0 +1,110 @@
...
+    tableOps.setProperty(tableNames[1], Property.TABLE_DURABILITY.getKey(), "flush");
+    tableOps.setProperty(tableNames[2], Property.TABLE_DURABILITY.getKey(), "log");
+    tableOps.setProperty(tableNames[3], Property.TABLE_DURABILITY.getKey(), "none");
+    return tableNames;
+  }
+
+  private void createTable(AccumuloClient c, String tableName) throws Exception {
+    c.tableOperations().create(tableName);
+  }
+
```
keith-turner: It would be nice to set the table props at table creation time.  The following is one possible way to do this.  Also, the getUniqueNames function could be dropped.  That functions makes sense for ITs, but not here.

```java
  private String[] init(AccumuloClient c) throws Exception {
    String[] tableNames = new String[]{"syncT","flushT","logT","noneT"};
    
    createTable(c, tableNames[0], "sync");
    createTable(c, tableNames[1], "flush");
    createTable(c, tableNames[2], "log");
    createTable(c, tableNames[3], "none");
    
    return tableNames;
  }

  private void createTable(AccumuloClient c, String table, String durability) throws Exception {
    NewTableConfiguration ntc  = new NewTableConfiguration();
    ntc.setProperties(Collections.singletonMap("table.durability", durability));
    c.tableOperations().create(table, ntc);
  }
```

src/main/java/org/apache/accumulo/testing/performance/tests/DurabilityWriteSpeedPT.java
```diff
@@ -0,0 +1,110 @@
...
+    Report.Builder reportBuilder = Report.builder();
+    reportBuilder.id("durability");
+    reportBuilder.description("Compares writes speeds at different durability levels");
+    try (AccumuloClient client = env.getClient()) {
+      TableOperations tableOps = client.tableOperations();
+      String[] tableNames = init(client);
+      // write some gunk, delete the table to keep that table from messing with the performance
+      // numbers of successive calls
+      // sync
+      long t0 = writeSome(reportBuilder, client, tableNames[0], N, "Sync");
```
keith-turner: Looking at this, it seems like this test could be simplified with a loop... something like :

```java
  for(String durability : new String[]{"sync","flush","log","none"}){
     String tableName = durability+"T";
      createTable(c, tableName, durability);
      long median = writeSome(reportBuilder, client, tableName, N, durability);
      tableOps.delete(tableName);
      reportBuilder.result( durability+" Median", median, "Median time result for "+durability);
  }
```

Manno15:
Thanks for the recommendations. I will implement these changes and re-push.

Manno15:
I noticed some changes in results, especially for sync, and by a fair margin so I am doing some digging as to why.

Issue #30, opened 2018-07-25
Add performance test for conditional mutations


keith-turner:
Could possibly use the following for inspiration.

https://github.com/keith-turner/CMPT

elinaawise:
I'm looking into this.

PR #75, opened 2019-05-24
Fix #67: Test group commit performance

### LINKED ISSUE #67: Add performance test for group commit
When many clients write data to the same tablet server at around the same time their data should be synced to the write ahead log as a group.  If group commit is not working properly it can cause performance problem for many clients that will not be seen for a single client.  It would be nice to have a performance test that specifically checks for this.

I created a project to do this in the past.  Not sure what state it is in.  

  https://github.com/keith-turner/mutslam

Would be nice to test group commit performance for mutations and conditional mutations.

elinaawise:
I migrated Keith's mutslam project, added Report logging and removed any deprecated code.  Issue #30 covers conditional mutations, which I plan to work on next.


src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,243 @@
...
+
+    double rate = totalNumMutations / (sum / (double) wasks.size());
+
+    report.info("time", sum / (double) wasks.size(),
+        "Time it took the task to run in milliseconds");
+    report.info("threads", numThreads, "Number of threads");
+    report.info("batch", numToBatch, "Number of batchs");
+    report.info("mutations", totalNumMutations, "Total number of mutations");
+    report.info("rate", rate, "Mutations per millisecond");
+
```
keith-turner: It seems like it will be hard to tell these apart in the test output because of the same id.  Could add a one up counter to the ids like :

```java
// testRun is a counter that is passed to method and incremented for every test
report.info("time"+testRun, sum / (double) wasks.size(),
￼         "Time it took the task to run in milliseconds");
￼     report.info("threads"+testRun, numThreads, "Number of threads");
```

It would also be useful to include `numToWrite`, `walog`, and `useSharedBW` in the report

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  @Override
+  public SystemConfiguration getSystemConfig() {
+
+    Map<String,String> siteCfg = new HashMap<>();
+
+    siteCfg.put(Property.TSERV_MINTHREADS.getKey(), "128");
+    siteCfg.put(Property.TABLE_DURABILITY.getKey(), "log");
```
keith-turner: Why set this durability setting?

elinaawise: I was following your mutslam code in which the Generator.java class had TABLE_WALOG_ENABLED set in runTest(...).  The TABLE_WALOG_ENABLED has been replaced by TABLE_DURABILITY, which default is "sync".

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
elinaawise: I was thinking the same thing actually.  I'll update it.

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,243 @@
...
+
+  private void runBatch(Environment env, Report.Builder report, int numThreads, int numToBatch,
+      boolean useSharedBW, boolean walog) throws Exception {
+
+    String tableName = "mutslam";
+
+    try {
+      env.getClient().tableOperations().create(tableName);
+      if (!walog) {
+        env.getClient().tableOperations().setProperty(tableName, Property.TABLE_DURABILITY.getKey(),
```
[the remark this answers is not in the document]
elinaawise: k. Looking into it.  I think I saw that being done elsewhere.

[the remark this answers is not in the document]
elinaawise: > k. Looking into it. I think I saw that being done elsewhere.

Actually, can you clarify what you mean?  I cannot find the code I thought I saw.

elinaawise:
> I think this entire test needs to be reorganized in order to produce numbers that are comparable. Instead of reporting on the three different test multiple times, I think it would be better to report stats for each test type once. This could be follow on work.

I modified the test to run a loop of 6 iterations and report on threads and batches using a single batch.  I did not see a difference in results when running a 3 iterations loop (warm up) prior to the 6 iterations, so I removed the "warm up" loops.

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    long sum = 0;
+    for (WriteTask writeTask : wasks) {
+      sum += writeTask.getTime();
+    }
+
+    int totalNumMutations = numToWrite * numThreads * numToBatch;
+
+    ArrayList<Object> returnValues = new ArrayList<>();
```
keith-turner: Would be a little bit cleaner to return a simple POJO rather than `ArrayList<Object>`

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,234 @@
...
+    returnValues.add(sum / (double) wasks.size()); // time
+    returnValues.add(totalNumMutations);
+    returnValues.add(totalNumMutations / (sum / (double) wasks.size())); // rate
+
+    // System.out.printf(
```
keith-turner: Could remove this

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,234 @@
...
+      // This tests a single thread writing a different batch sizes of mutations,
+      // flushing after each batch. Group commit should approach these times for the same number
+      // mutations.
+      batchValues.clear();
+      batchValues = runBatch(env, report, 1, numThreads);
+
+      report.info("batchTime" + i,
+          new Double(new DecimalFormat("#0.00").format(batchValues.get(0))),
+          "Time it took the task to run in milliseconds");
+      report.info("threads" + i, i, "Number of threads");
```
keith-turner: This will have the same name as key from prev loop.

src/main/java/org/apache/accumulo/testing/performance/tests/GroupCommitPT.java
```diff
@@ -0,0 +1,234 @@
...
+          "Total number of mutations");
+      report.result("batchRate" + i,
+          new Double(new DecimalFormat("#0.00").format(batchValues.get(2))),
+          "Total mutations per millisecond");
+    }
+
+    return report.build();
+  }
+
+  private ArrayList<Object> runBatch(Environment env, Report.Builder report, int numThreads,
```
keith-turner: The report is passed in, but is it used?

elinaawise: Nope, sure isn't.  Thanks for catching that.

Manno15:
@keith-turner Do you know the status of this pull request or if it is ready to be merged?

keith-turner:
> @keith-turner Do you know the status of this pull request or if it is ready to be merged?

I have wanted to look at these performance test but have not had time as I am focusing on things related to getting 2.0 released.   I will look at them, but it may be after we release 2.0.

Manno15:
That's completely understandable, I was just going back through some of the older pull requests. I can take care of anything that needs to be changed or edited.

ctubbsii:
@keith-turner Just revisiting this open PR. Should we go ahead and merge this? If somebody picks them up, later, more work can be done on them. In the meantime, I can't imagine they are going to cause problems, since they are independent of the other test code in the repo.

Manno15:
I am going through this to make sure it works properly. I will handle any corrections that need to be made.

Manno15:
Looking through this, there are several changes needed to be made. What is the best way to make those? This to be merged and then for me to make a pull request against it? I am also having some trouble connecting to zookeeper with the performance test suite so I am looking into that so I can properly test this.

Manno15:
It would probably be best if @keith-turner (if you have time) looks over the test functionality itself to see if it meets his original intention or if I need to add anything to it. I have made some of his other requested changes and I am editing the report so it makes more sense.

keith-turner:
@Manno15 I am taking a stab at finishing this.

keith-turner:
I opened #125 as a follow on to finish this up.

PR #77, opened 2019-06-03
Remove Snappy Property for CreateTable

Manno15:
Snappy seems to not come by default with the latest version of Hadoop. So removed that line when it comes to creating a table. It will now use the default which is gz.


keith-turner:
The changes look good.  @Manno15 did you try running continuous ingest with these changes?

Manno15:
Yep, everything worked. Not sure if the different compression types have better or worse performance metrics though.

PR #79, opened 2019-06-07
Issue30: Test performance of Conditional Mutations

elinaawise:


elinaawise:
To simplify the data being reported, I commented out all reporting except the average rate values.

ctubbsii:
Please consider changing the title of this PR to be more descriptive about what it intends to do. Also please consider adding a description. Doing so will give context for any reviewers.

Manno15:
There appears to be some comments that need to be deleted but beyond that it appears to be ready to be merged. Thoughts @ctubbsii @keith-turner ?

ctubbsii:
> There appears to be some comments that need to be deleted but beyond that it appears to be ready to be merged. Thoughts @ctubbsii @keith-turner ?

Yes, deleting the commented out code and/or dead code would be a nice cleanup before merging. Could merge as-is, if you wanted to do that follow-on, @Manno15

Manno15:
Doing it as a follow-on would probably be easier on my end but I am okay with whichever you prefer.

ctubbsii:
Merged. Thanks, @elinaawise ; @Manno15 , feel free to perform any follow-on work in a separate PR to clean up/improve as you see fit.

PR #81, opened 2019-06-17
Add new garbage collection simulation test

keith-turner:


keith-turner:
I left this running overnight on a single node.  For the test I ran 10 concurrent generators.  During that time 285 million items were deleted by the collector.  I shut everything down and ran verify and the referential integrity was ok.

ivakegg:
+1

PR #85, opened 2019-06-21
update bulk docs

keith-turner:


docs/bulk-test.md
```diff
@@ -24,15 +24,25 @@ for i in $(seq 1 10); do
...
+done
+
 ./bin/cingest verify
 ```
 
 Bulk ingest could be run concurrently with live ingest into the same table.  It
 could also be run while the agitator is running.
 
+After bulk imports complete, could run the following commands in the Accumulo shell
+to see if there are any blips or load markers.  There should not be any.
```
milleruntime: Should define what a BLIP is since technically it is an acronym.

PR #90, opened 2019-07-11
Move performance tests to accumulo-testing #1200

lbschanno:
Move ManySplitIT as HighSplitCreationPT.
Move BalanceFasterIT as SplitBalancingPT.
Move DeleteTableDuringSplitIT as TableDeletionDuringSplitPT.
Move RollWALPerformanceIT as RollWALPT.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
@@ -1,0 +1,72 @@
+package org.apache.accumulo.testing.performance.tests;
+
+import java.nio.charset.StandardCharsets;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.metadata.MetadataTable;
+import org.apache.accumulo.testing.performance.Environment;
+import org.apache.accumulo.testing.performance.PerformanceTest;
+import org.apache.accumulo.testing.performance.Report;
+import org.apache.accumulo.testing.performance.SystemConfiguration;
+import org.apache.hadoop.io.Text;
+
+public class HighSplitCreationPT implements PerformanceTest {
+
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
```diff
@@ -0,0 +1,72 @@
...
+    long start = System.currentTimeMillis();
+    client.tableOperations().addSplits(TABLE_NAME, splits);
+    long totalTime = System.currentTimeMillis() - start;
+    double splitsPerSecond = NUM_SPLITS / (totalTime / ONE_SECOND);
+
+    reportBuilder.parameter("splits_per_second", splitsPerSecond,
+        "The average number of splits created per second.");
+    reportBuilder.parameter("meets_performance_threshold",
+        splitsPerSecond > MIN_REQUIRED_SPLITS_PER_SECOND,
+        "Whether or not the average split creation meets or exceeds the minimum number of splits per second.");
```
Manno15: `reportBuilder.parameter` should be replaced with `reportBuilder.result` for the resulting value of the performance test. Results are written in a different section of the json file that gets written.

Manno15: The same should be done for the results of the other tests as well.

lbschanno: @Manno15 thanks for the input, I'll update the code to do so.

lbschanno: Changes made.

milleruntime:
@Manno15 @lbschanno where you able to run these on a cluster using the performance scripts?  I am not sure of the state of the performance framework but I think these are OK to merge.  It is just difficult to tell what you changed after moving over.

milleruntime:
I will go ahead and merge these tests.  There can be follow up work if any are failing or not working.

PR #91, opened 2019-07-11
Improve scripts

milleruntime:
* Made scripts all call ./bin/build
* Added check for env vars in env.sh
* Print out where the env vars are coming from

### DIFF
--- a/bin/agitator
+++ b/bin/agitator
@@ -17,6 +17,7 @@
 
 bin_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 at_home=$( cd "$( dirname "$bin_dir" )" && pwd )
+source "${bin_dir}/build"
 
 function print_usage() {
   cat <<EOF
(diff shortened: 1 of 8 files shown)


bin/agitator
```diff
@@ -17,6 +17,7 @@
 
 bin_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
 at_home=$( cd "$( dirname "$bin_dir" )" && pwd )
+source "${bin_dir}/build"
```
ctubbsii: Do we want to source this file, or execute it? (source will insert any variables set in the build command into the current execution environment, vs. execute, which will treat them as local to the build command).

milleruntime: build is sourcing $at_home/conf/env.sh so I assume a source is required?

ctubbsii: Typically, sourcing an env file makes sense because the the whole point of an env script is to create environment variables that the caller wants to include in its own environment.

I don't know enough about the build command to know if it is intended to behave the same way as an env script or if it's supposed to be merely executing and returning. If it's the latter, you don't need to source it.

milleruntime: The build script is basically just doing a ```source env.sh``` and then env.sh will build the shaded jar.

ctubbsii: In that case, it needs to be sourced, but the build name is just confusing, since all it's doing is sourcing the env script.

PR #93, opened 2019-07-15
Fixes #59 Create herding performance test

### LINKED ISSUE #59: Create herding performance test
I wrote the following test while working on apache/accumulo#990.  This could be turned into a performance test.

```java
public class HerdTest {

  private static final byte[] E = new byte[] {};
  private static final byte[] FAM = "pinky".getBytes();

  private static final int NUM_ROWS = 1_000_000;
  private static final int NUM_COLS = 10;

  public static void main(String[] args) throws Exception {

    Connector conn = CmdUtil.getConnector();

    if (!conn.tableOperations().exists("herd")) {
      conn.tableOperations().create("herd", new NewTableConfiguration().setProperties(
          Collections.singletonMap(Property.TABLE_BLOCKCACHE_ENABLED.getKey(), "true")));
      write(conn);
      conn.tableOperations().flush("herd", null, null, true);
    }

    testHerd(conn, 32);
  }

  private static void testHerd(Connector conn, int nt)
      throws InterruptedException, ExecutionException {
    ExecutorService tp = Executors.newFixedThreadPool(nt);
    final CyclicBarrier cb = new CyclicBarrier(nt);

    long t1 = System.currentTimeMillis();
    List<Future<?>> futures = new ArrayList<>();
    for (int i = 0; i < nt; i++) {
      Future<?> f = tp.submit(new Runnable() {
        @Override
        public void run() {
          try {
            Scanner scanner = conn.createScanner("herd", Authorizations.EMPTY);

            for (int i = 0; i < 1000; i++) {

              // System.out.println(Thread.currentThread().getId()+" "+i);

              cb.await();

              byte[] row = FastFormat.toZeroPaddedString(i * 1000, 8, 16, E);

              scanner.setRange(Range.exact(new String(row)));
              for (Entry<Key,Value> entry : scanner) {

              }
            }
          } catch (Exception e) {
            e.printStackTrace();
          }
        }
      });
      futures.add(f);
    }

    for (Future<?> future : futures) {
      future.get();
    }

    long t2 = System.currentTimeMillis();

    System.out.println(t2 - t1);

    // scanner.close();
    tp.shutdown();
  }

  private static void write(Connector conn) throws Exception {

    try (BatchWriter bw = conn.createBatchWriter("herd", new BatchWriterConfig())) {
      Random rand = new Random();

      for (int r = 0; r < NUM_ROWS; r++) {
        byte[] row = FastFormat.toZeroPaddedString(r, 8, 16, E);
        Mutation m = new Mutation(row);
        for (int c = 0; c < NUM_COLS; c++) {
          byte[] qual = FastFormat.toZeroPaddedString(c, 4, 16, E);

          byte[] val = new byte[32];
          rand.nextBytes(val);

          m.put(FAM, qual, val);
        }

        bw.addMutation(m);
      }
    }
  }
}
```

lbschanno:
Add HeardingPT.java in order to test herding performance.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/HerdingPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/HerdingPT.java
@@ -1,0 +1,165 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.accumulo.testing.performance.tests;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.CyclicBarrier;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.client.AccumuloException;
+import org.apache.accumulo.core.client.AccumuloSecurityException;
+import org.apache.accumulo.core.client.BatchWriter;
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/performance/tests/HerdingPT.java
```diff
@@ -0,0 +1,165 @@
...
+        .parameter("table_name", TABLE_NAME, "The name of the table.")
+        .parameter("num_threads", NUM_THREADS, "The number of threads.")
+        .parameter("num_load_attempts", NUM_LOAD_ATTEMPTS,
+            "The number of times the threads will attempt to load the same block.")
+        .parameter("num_rows", NUM_ROWS, "The number of rows that will be loaded into the table.");
+
+    final AccumuloClient client = env.getClient();
+    initTable(client);
+    long herdTime = getHerdingDuration(client);
+    reportBuilder.parameter("herd_time", herdTime, "The time (in ms) it took herding to complete.");
```
Manno15: `reportBuilder.parameter` should be replaced with `reportBuilder.result` for the resulting value of the performance test

lbschanno: Code updated.

Manno15:
It appears the build for this needs to be re-run. It builds locally with the same command.

PR #96, opened 2019-07-18
Resolves #78 Script to automate EC2 

### LINKED ISSUE #78: Create script that automates setting up EC2 test cluster
When testing Accumulo I often go through the following task manually.

 * Build snapshot version of Accumulo locally and copy it to Muchos dir
 * Use Muchos to setup a cluster with that snapshot tarball
 * After Muchos sets up the cluster :
    * git clone and  mvn install the same snapshot version of Accumulo
    * git clone accumulo testing repo and build it.

It may be nice to have script that does this.  The argument for the script would be the following :
 * URL for Accumuo git repo and a branch in that repo
 * URL for Accumuo testing repo and a branch in that repo
 * Local dir where muchos is setup.

This script would setup the EC2 cluster with the version of Accumulo from the git repo.  It would also setup accumulo testing on the cluster from the git repo.

rcarterjr:
This script implements the initial features requested. The script requires three arguments in order: 

1. URL for Accumuo git repo and a branch in that repo
2. URL for Accumuo testing repo and a branch in that repo
3. Local dir where muchos is setup.

I will continue working on this and adding the long term goals.


bin/automateEC2
```diff
@@ -0,0 +1,77 @@
...
+	# installs java-11 and makes it the default for mvn
+	ssh $CLUSTERUSER@$PROXYIP "sudo yum -y install java-11-openjdk-devel &"
+	ssh $CLUSTERUSER@$PROXYIP "sed -i '/export JAVA_HOME/d' ~/.bashrc"
+	ssh $CLUSTERUSER@$PROXYIP "sed -i '/# .bashrc/a export JAVA_HOME=/usr/lib/jvm/java-11-openjdk' ~/.bashrc"
+	ssh $CLUSTERUSER@$PROXYIP "source ~/.bashrc"
+
+	# clones and builds accumulo and  accumulo-testing to EC2
+	ssh $CLUSTERUSER@$PROXYIP "git clone --single-branch --branch 2.0 --depth=1 $ACCUMULO_REPO &"
+	ssh $CLUSTERUSER@$PROXYIP "cd accumulo && $M2/mvn clean install -PskipQA && cd .. &"
+	ssh $CLUSTERUSER@$PROXYIP "git clone ACCUMULO_TESTING_REPO &"
```
ctubbsii: Should that have a `$`?

rcarterjr: Yes, thanks I just added that variable. Adding now

bin/automateEC2
```diff
@@ -0,0 +1,77 @@
...
+		./bin/muchos terminate -c $USER-cluster
+                break
+        fi
+
+	CLUSTERUSER=`./bin/muchos config -p cluster_user`
+	PROXYIP=`./bin/muchos config -p proxy.public.ip`
+	M2='/home/centos/install/apache-maven*/bin'
+	
+	# installs java-11 and makes it the default for mvn
+	ssh $CLUSTERUSER@$PROXYIP "sudo yum -y install java-11-openjdk-devel &"
```
keith-turner: I opened apache/fluo-muchos#265 to do this in muhcos

milleruntime:
@rcarterjr what's the status of this PR?  It looks OK maybe just needs someone to test it out on a cluster?

bin/automateEC2
```diff
@@ -0,0 +1,77 @@
...
+	ssh $CLUSTERUSER@$PROXYIP "sed -i '/export JAVA_HOME/d' ~/.bashrc"
+	ssh $CLUSTERUSER@$PROXYIP "sed -i '/# .bashrc/a export JAVA_HOME=/usr/lib/jvm/java-11-openjdk' ~/.bashrc"
+	ssh $CLUSTERUSER@$PROXYIP "source ~/.bashrc"
```
milleruntime: It looks like this can be removed and handled by Muchos with @keith-turner 's change.

rcarterjr: That's correct, I removed it on my most recent commit but I haven't pushed it. Once I test it further I will push, but I don't get back to the office until next week.

bin/automateEC2
```diff
@@ -0,0 +1,77 @@
...
+	mvn clean package -DskipFormat -PskipQA
+	
+	# copies the tarball to the given muchos directory
+	cp ./assemble/target/*.gz $MUCHOS_REPO/conf/upload/
+	if [ $? -eq 0 ]; then
+		echo "Accumulo tarball copied to fluo-muchos"
+		rm -rf ~/accumulotmp
+	else
+		break
+		rm -rf ~/accumulotmp
```
milleruntime: The line after the break is unreachable.  I think you meant to before the break.

rcarterjr:
This script now takes the args.txt as its only argument with the following inputs on their own line in the following order:
- accumulo git repository 
- accumulo branch
- accumuluo-testing git repository
- accumulo-testing branch
- fluo-muchos git repository
- fluo-muchos branch
- path/to/muchos.props

I will work on creating environment variables instead of having the text file and iterating through as discussed.

rcarterjr:
@milleruntime @keith-turner here are my latest changes. Let me know what you think.

test/automation/cluster_props.sh
```diff
@@ -0,0 +1,28 @@
...
+export ACCUMULO_BRANCH=master
+
+export ACCUMULO_TESTING_REPO=https://github.com/apache/accumulo-testing.git
+export ACCUMULO_TESTING_BRANCH=master
+
+export FLUO_MUCHOS_REPO=https://github.com/apache/fluo-muchos
+export FLUO_MUCHOS_BRANCH=master
+
+# path/to/muchos.props
+export MUCHOS_PROPS=
```
milleruntime: File looks good, just need a default value for MUCHOS_PROPS

```suggestion
export MUCHOS_PROPS=${MUCHOS_PROPS:-/path/to/muchos/props}
```
If you set it using this syntax the value can come from your env if set, otherwise default to a non existent path.  This way a user can just set the value in the env or modify this value.

rcarterjr: Made those changes, thanks for the feedback.

milleruntime:
I will merge this up but we can continue to test and make improvements.

milleruntime:
@rcarterjr as another follow on task, you could run ShellCheck on the script: https://github.com/koalaman/shellcheck.

rcarterjr:
Thanks Mike, that's a neat tool.

And I will write up a readme now Keith.

keith-turner:
> And I will write up a readme now Keith.

When you do that I will try to follow the readme and run this thing.  So far I have only reviewed the code script, I have not tried to run it.

PR #99, opened 2019-07-22
Replace Thrift types with TableNotFoundException

milleruntime:


src/main/java/org/apache/accumulo/testing/randomwalk/concurrent/Config.java
```diff
@@ -153,10 +150,8 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
           env.getAccumuloClient().namespaceOperations().setProperty(namespace, property.getKey(),
               property.getDefaultValue());
         } catch (AccumuloException ex) {
-          if (ex.getCause() instanceof ThriftTableOperationException) {
-            ThriftTableOperationException ttoe = (ThriftTableOperationException) ex.getCause();
-            if (ttoe.type == TableOperationExceptionType.NAMESPACE_NOTFOUND)
-              return;
+          if (ex.getCause() instanceof TableNotFoundException) {
```
ctubbsii: Should this be NamespaceNotFoundException?

milleruntime: I think so good catch.  Fixed in 6a87136

src/main/java/org/apache/accumulo/testing/randomwalk/concurrent/Config.java
```diff
@@ -228,10 +222,8 @@ private void changeNamespaceSetting(RandomDataGenerator random, State state, Ran
       env.getAccumuloClient().namespaceOperations().setProperty(namespace,
           setting.property.getKey(), "" + newValue);
     } catch (AccumuloException ex) {
-      if (ex.getCause() instanceof ThriftTableOperationException) {
-        ThriftTableOperationException ttoe = (ThriftTableOperationException) ex.getCause();
-        if (ttoe.type == TableOperationExceptionType.NAMESPACE_NOTFOUND)
-          return;
+      if (ex.getCause() instanceof TableNotFoundException) {
```
ctubbsii: This one also seems to be looking for a NamespaceNotFoundException

milleruntime: Yup, nice catch.

PR #100, opened 2019-07-22
Drop example files

milleruntime:


milleruntime:
>  However, I think there's still some logic that looks for the files with ".example", which should also be updated:
> 
> ```shell
> grep -R "[.]example" *
> ```

Good call.  I will remove references to the examples as well.

milleruntime:
There are a few example files for other tests that I am not sure if we need them or not: 
```
./test/agitator/agitator.ini.example
./test/agitator/hosts.example
./test/scalability/conf/Ingest.conf.example
./test/scalability/conf/site.conf.example
./test/stress/stress-env.sh.example
```

ctubbsii:
> There are a few example files for other tests that I am not sure if we need them or not:
> 
> ```
> ./test/agitator/agitator.ini.example
> ./test/agitator/hosts.example
> ./test/scalability/conf/Ingest.conf.example
> ./test/scalability/conf/site.conf.example
> ./test/stress/stress-env.sh.example
> ```

Probably best to leave them for now, since their unrelated to these particular (most commonly used) files.

bin/build
```diff
@@ -18,10 +18,6 @@
...
-if [ -f "$at_home/conf/env.sh" ]; then
-  conf="$at_home/conf/env.sh"
-else
-  conf="$at_home/conf/env.sh.example"
-fi
-echo "Using $conf for setup"
-source $conf
+env_sh="$at_home/conf/env.sh"
+echo "Using $env_sh for setup"
+source $env_sh
```
keith-turner: What happens if the files does not exist?

ctubbsii: The file always exists unless you deleted it from the repo after checking it out.

PR #101, opened 2019-07-22
BulkIngest CI improvements

ctubbsii:
* Add job name
* Seed Random with strong seed
* Fix linked list to chain adjacent elements, rather than skip over
  adjacent elements to link to the next element

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
@@ -49,13 +49,14 @@ public class BulkIngest extends Configured implements Tool {
 
   @Override
   public int run(String[] args) throws Exception {
+    String ingestInstanceId = UUID.randomUUID().toString();
+
     Job job = Job.getInstance(getConf());
+    job.setJobName("BulkIngest_" + ingestInstanceId);
     job.setJarByClass(BulkIngest.class);
     // very important to prevent guava conflicts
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
```diff
@@ -49,14 +49,15 @@
 
   @Override
   public int run(String[] args) throws Exception {
+    String ingestInstanceId = UUID.randomUUID().toString();
+
     Job job = Job.getInstance(getConf());
+    job.setJobName("BulkIngest_" + ingestInstanceId);
```
keith-turner: I like having the uuid in the name because it could be useful for debugging if data loss were to happen.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousInputFormat.java
```diff
@@ -158,9 +165,8 @@ public boolean nextKeyValue() {
 
         if (nodeCount < numNodes) {
           CRC32 cksum = checksum ? new CRC32() : null;
-          byte[] prevRow = prevKey != null ? prevKey.getRowData().toArray() : null;
-
           prevKey = currKey;
+          byte[] prevRow = prevKey != null ? prevKey.getRowData().toArray() : null;
```
keith-turner: Because of this I was ending up with double the unreferenced counts than I expected when I ran the verification map reduce job.  I will update the docs about expected counts to be more specific.

PR #105, opened 2019-07-24
Clean up and Minor Changes to ConditionalMutationsPT

Manno15:
This is in reference to PR #79. Cleaned up unused and unnecessary code.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
@@ -115,8 +115,8 @@ public class ConditionalMutationsPT implements PerformanceTest {
       cmuts.add(cm);
     }
 
-    // long t1 = System.currentTimeMillis();
     long t1 = System.nanoTime();
+
     int count = 0;
     Iterator<ConditionalWriter.Result> results = cw.write(cmuts.iterator());
     while (results.hasNext()) {
@@ -134,7 +134,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
+    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
   }
 
   private static void runRandomizeConditionalMutationsTest(Environment env, String tableName,
@@ -198,8 +198,8 @@ public class ConditionalMutationsPT implements PerformanceTest {
     }
     cmuts.add(cm);
 
-    // long t1 = System.currentTimeMillis();
     long t1 = System.nanoTime();
+
     int count = 0;
     Iterator<ConditionalWriter.Result> results = cw.write(cmuts.iterator());
     while (results.hasNext()) {
@@ -217,7 +217,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
+    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
   }
 
   private static void runRandomizeBatchScanAndWriteTest(Environment env, String tableName,
@@ -278,8 +278,8 @@ public class ConditionalMutationsPT implements PerformanceTest {
     bw.addMutation(cm);
     bw.flush();
 
-    // long t1 = System.currentTimeMillis();
     long t1 = System.nanoTime();
+
     bs.setRanges(ranges);
 
     int count = 0;
@@ -292,7 +292,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
+    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
   }
 
   private static void runSetBlockSizeTest(Environment env, String tableName,
@@ -422,8 +422,8 @@ public class ConditionalMutationsPT implements PerformanceTest {
       cmuts.add(cm);
     }
 
-    // long t1 = System.currentTimeMillis();
     long t1 = System.nanoTime();
+
     int count = 0;
 
     Iterator<ConditionalWriter.Result> results = cw.write(cmuts.iterator());
@@ -442,6 +442,6 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 30000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
+    return 30000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
   }
 }
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
@@ -52,7 +52,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     String tableName = "foo";
 
     Report.Builder reportBuilder = Report.builder();
-    reportBuilder.id("condmut");
+    reportBuilder.id("Conditional Mutations");
     reportBuilder.description("Runs conditional mutations tests with and without randomization,"
         + "setting of block size, and with and without randomization of batch writing/reading.");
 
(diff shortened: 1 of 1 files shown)


Manno15:
@ctubbsii are you familiar with the error that occurred with travis? It appears it failed during the installation of JAVA phase. I am unsure as to why

ctubbsii:
I pushed a change to the travis config. Apparently Oracle JDK 8 can't be downloaded and installed anymore. So, I switched to openjdk8. At some point, we'll update to 11 anyway, so this is an interim fix.

ctubbsii:
You may have to rebase and force-push, or just add a new commit to force it to rebuild.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -76,35 +76,31 @@ private static void runConditionalMutationsTest(Environment env, String tableNam
     ConditionalWriter cw = env.getClient().createConditionalWriter(tableName,
         new ConditionalWriterConfig());
 
-    conditionalMutationsTime(cw, null, "0", reportBuilder);
+    conditionalMutationsTime(cw, null);
 
     double rateSum = 0.0;
     for (int i = 1; i < 20; i++) {
-      rateSum += conditionalMutationsTime(cw, (long) i, ((Integer) i).toString(), reportBuilder);
+      rateSum += conditionalMutationsTime(cw, (long) i);
```
ctubbsii: Could just change the type of `i` to `long`... since it's going to be cast anyway.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -138,12 +134,6 @@ public static double conditionalMutationsTime(ConditionalWriter cw, Long seq, St
 
     double rate = 10000 / ((t2 - t1) / 1000.0);
```
ctubbsii: nanoTime, rather than clock time, is more appropriate for differences. Could also convert the time explicitly using TimeUnit functions in the divisor rather than having a divisor that is itself a division problem.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -263,53 +231,41 @@ private static void runRandomizeBatchScanAndWriteTest(Environment env, String ta
...
-
-    // System.out.printf("rate avg : %6.2f \n", rateSum / 20);
+    reportBuilder.result("avgRate: 20-39",
+        new Double(new DecimalFormat("#0.00").format(rateSum / 20)),
+        "RandomizeBatchScanAndWriteTest: average rate (in seconds) to write and scan sequence 20-39 post flush");
   }
 
-  private static double randomizeBatchWriteAndScanTime(BatchWriter bw, BatchScanner bs, Long seq,
-      String testRun, Report.Builder reportBuilder, boolean randomize) throws Exception {
+  private static double randomizeBatchWriteAndScanTime(BatchWriter bw, BatchScanner bs, Long seq)
```
ctubbsii: Should probably use lower-case `long` instead of `Long`, if `null` has no special meaning. Otherwise, this is creating unnecessary boxing/unboxing.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -263,53 +231,41 @@ private static void runRandomizeBatchScanAndWriteTest(Environment env, String ta
...
+      throws Exception {
 
     ArrayList<Range> ranges = new ArrayList<>();
-
     Mutation cm = new Mutation("r01");
-
     ArrayList<Integer> ints = new ArrayList<>(10000);
 
     for (int i = 0; i < 10000; i++) {
```
ctubbsii: 10000 is used a lot. Should it be a constant?

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
Manno15: I agree, I am not fully sure on the initial reasoning behind using `Long`. I made the change and replaced the null's with 0 to keep her code logic mostly the same.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
Manno15: Possibly, 1000 and 100000 are also used quite a bit. Could add constants for all of them. I also see some room to had parameters to the report I hadn't noticed before based off of some of these numbers.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -130,11 +131,10 @@ public static double conditionalMutationsTime(ConditionalWriter cw, Long seq) th
     if (cmuts.size() != count) {
       throw new RuntimeException();
     }
-    long t2 = System.currentTimeMillis();
 
-    double rate = 10000 / ((t2 - t1) / 1000.0);
+    long t2 = System.nanoTime();
 
-    return rate;
+    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
```
ctubbsii: Should be `NANOSECONDS.toSeconds`, but otherwise, this looks much cleaner.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -214,9 +214,10 @@ private static double randomizeConditionalMutationsTime(ConditionalWriter cw, Lo
     if (cmuts.size() != count) {
       throw new RuntimeException();
     }
-    long t2 = System.currentTimeMillis();
 
-    return 10000 / ((t2 - t1) / 1000.0);
+    long t2 = System.nanoTime();
+
+    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
```
ctubbsii: Also should be `NANOSECONDS`

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -289,9 +290,9 @@ private static double randomizeBatchWriteAndScanTime(BatchWriter bw, BatchScanne
       throw new RuntimeException("count = " + count);
     }
 
-    long t2 = System.currentTimeMillis();
+    long t2 = System.nanoTime();
 
-    return 10000 / ((t2 - t1) / 1000.0);
+    return 10000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
```
ctubbsii: Another one. :smiley_cat:

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -434,10 +440,8 @@ private static double setBlockSizeTime(ConditionalWriter cw, int numRows, int nu
       throw new RuntimeException();
     }
 
-    long t2 = System.currentTimeMillis();
-
-    double rate = 30000 / ((t2 - t1) / 1000.0);
+    long t2 = System.nanoTime();
 
-    return rate;
+    return 30000 / TimeUnit.MICROSECONDS.toSeconds(t2 - t1);
```
ctubbsii: Last one. :1234:

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
Manno15: I did nanoseconds to seconds first but it was returning 0 when the conversion was applied. Microseconds was the only one that returned nonzero number so I rolled with it. The results are comparable to the previous ones.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
Manno15: Same reasoning for the rest of these too.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
ctubbsii: NANOSECONDS is the correct units here. But, you'll have to cast either the numerator or denominator to a double to get a non-zero rate. Something like:

```java
return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2-t1);
```
or
```java
return 10000 / (double) TimeUnit.NANOSECONDS.toSeconds(t2-t1);
```

Manno15: ah, that makes a lot of sense.I figured it was something to do with the return type but didn't know just adding the .0 at the end of the numerator would fix that. I will fix these tomorrow, thank you.

Manno15:
I made the changes you requested. Sorry it took so long, I've been dealing with some issues running any of the performance tests for some reason. Due to that, I wasn't able to check the final test results after changing to nanoseconds. I pushed anyways since I wanted to get it out before the weekend.

PR #108, opened 2019-08-08
Removed SNAPSHOT from version

Manno15:
The first version line still contained SNAPSHOT in it. The Jar file would in turn contain SNAPSHOT in its name.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
   </parent>
   <groupId>org.apache.accumulo</groupId>
   <artifactId>accumulo-testing</artifactId>
-  <version>2.0.0-SNAPSHOT</version>
+  <version>2.0.0</version>
   <name>Apache Accumulo Testing</name>
   <description>Testing tools for Apache Accumulo</description>
   <properties>


ctubbsii:
> The first version line still contained SNAPSHOT in it. The Jar file would in turn contain SNAPSHOT in its name.

Why is that a problem? No accumulo-testing artifact has been voted on for a release, and should not have a version which appears to be a release version.

Manno15:
oh, I think I was confused on what the line represented. My mistake. Will close this pull request then.

ctubbsii:
> oh, I think I was confused on what the line represented. My mistake. Will close this pull request then.

No problem. The similarity with the Accumulo version is because this repo was created from code exported from the main repo. Maybe it makes sense to change the version number to something like `0.0.1-SNAPSHOT` so it's more obvious that it's unrelated?

mjwall:
> 
> No problem. The similarity with the Accumulo version is because this repo was created from code exported from the main repo. Maybe it makes sense to change the version number to something like `0.0.1-SNAPSHOT` so it's more obvious that it's unrelated?

:+1: to changing the version

Manno15:
Yeah, as it stands, it looks like it is tied to whatever version accumulo is at.

Issue #114, opened 2019-10-06
Incorrect authorizations selected by continous scanners and walkers 


mandarinamdar:
Authorizations configured for continous scanners and walkers in accumulo-testing.properties in "test.ci.common.auths" incorrectly interpreted as single character each instead of | delimited. Like if  
test.ci.common.auths=SYS,HR,ADM|SYS,ADM is defined. Authorizations chosen are S,Y,S,H,R ... and so on instead of "SYS,HR,ADM" , "SYS,ADM" and so on. 

The issue looks to be in  authValue.split("|") call in ContinuousEnv.java. The code is not considering | as  a metacharacter in regex. It will require escape character to define | in the split function call. something like authValue.split("\\|") to handle the desired functionality correctly.

ctubbsii:
Fixed by #115

PR #115, opened 2019-10-06
Fixes issue #114, incorrect auths used by scanners and walkers

mandarinamdar:
Fixed authValue.split("|") call in ContinuousEnv.java.
It was not considering | as a metacharacter in regex used by split
added escape character "\\" in the split function call as authValue.split("\\|")

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousEnv.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousEnv.java
@@ -27,7 +27,7 @@ public class ContinuousEnv extends TestEnv {
         authList = Collections.singletonList(Authorizations.EMPTY);
       } else {
         authList = new ArrayList<>();
-        for (String a : authValue.split("|")) {
+        for (String a : authValue.split("\\|")) {
           authList.add(new Authorizations(a.split(",")));
         }
       }


src/main/java/org/apache/accumulo/testing/continuous/ContinuousEnv.java
```diff
@@ -27,7 +27,7 @@ public ContinuousEnv(String[] args) {
         authList = Collections.singletonList(Authorizations.EMPTY);
       } else {
         authList = new ArrayList<>();
-        for (String a : authValue.split("|")) {
+        for (String a : authValue.split("\\|")) {
```
ctubbsii: If the double-escaping ever gets annoying, you can also use the square bracket syntax to create a character class, rather than use the escape character, as in: `.split("[|]")`

keith-turner:
Thanks for the contribution @mandarinamdar.  If you would like to be added as a contributor on the people page, [please submit a PR](https://github.com/apache/accumulo-website/edit/master/pages/people.md)

ctubbsii:
Hi @mandarinamdar . Thanks for the contribution. If you're intending to contribute regularly, please also consider subscribing to our developer mailing list (details at https://accumulo.apache.org/contact-us/), and consider introducing yourself to the community.

PR #118, opened 2019-11-14
Adding Accumulo Availability Monitor

tushardhadiwal:
Initial code for an accumulo availability monitor which poses as a synthetic user and is doing scans of random values across various tablet servers and capturing timing information related to how long such scans take in order get a overall judgement of availability of all tablet servers.

Monitor.Java : This file includes core code for scan operation that randomly picks a tablet server, identifies its minimum and maximum rows range and picks out 'distance' (configurable through opts) number of rows to be scanned in one iteration.

MonitorOpts.java : This file is used to provide all the parameters which are taken by the Monitor Class. Interesting Options Include : Ability to set scanner sleep time , between iterations. Setting Batch Size that scanner uses internally. Setting number of iteration or continuous mode.

monitor script : This follows same pattern as cingest, rwalk, performance scripts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Tests are run using the following scripts in `bin/`:
   * `performance` - Runs performance test
   * `agitator` - Runs agitator
   * `gcs` - Runs garbage collection simultation
+  * `monitor` - Runs availability monitor probe
 
 Run the scripts without arguments to view usage.
 
@@ -43,6 +44,7 @@ run in Docker:
 
   * `cingest` - All applications can be run except `verify` & `moru` which launch a MapReduce job.
   * `rwalk` - All modules can be run.
+  * `monitor` - All modules can be run.
 
 1. To create the `accumulo-testing` docker image, make sure the following files exist in your clone:
 
@@ -237,6 +239,11 @@ test and produce json result files.
 
 There are some utilities for working with the json result files, run the `performance` script
 with no options to see them.
+
+## Availability Monitor 
+Monitor class aims at verifying availability of overall accumulo cluster by continually doing
+scans of random values across various tablet servers and capturing timing
+information related to how long such scans take. 
 
 ## Automated Cluster Testing
 See the [readme.md](/test/automation/README.md).
--- a/src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
+++ b/src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
@@ -1,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.testing.healthprobe;
+
+import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.Map.Entry;
+import java.util.Random;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.accumulo.core.client.Accumulo;
+import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.client.AccumuloException;
+import org.apache.accumulo.core.client.AccumuloSecurityException;
+import org.apache.accumulo.core.client.Scanner;
+import org.apache.accumulo.core.client.TableNotFoundException;
+import org.apache.accumulo.core.client.admin.TableOperations;
+import org.apache.accumulo.core.data.Key;
+import org.apache.accumulo.core.data.Range;
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.core.util.FastFormat;
+import org.apache.hadoop.io.Text;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+import com.google.common.collect.Lists;
+
+public class Monitor {
+
+  private static final Logger log = LoggerFactory.getLogger(Monitor.class);
+  private static final byte[] EMPTY_BYTES = new byte[0];
+  Random r = new Random();
+  static long distance = 100l;
+
+  public static void main(String[] args) throws Exception {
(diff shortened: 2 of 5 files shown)


src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  public static Range pickRange(TableOperations tops, String table, Random r)
+      throws TableNotFoundException, AccumuloSecurityException, AccumuloException {
+
+    ArrayList<Text> splits = Lists.newArrayList(tops.listSplits(table));
+    if (splits.isEmpty()) {
+      return new Range();
+    } else {
+      int index = r.nextInt(splits.size());
```
phrocker: What value does this provide over selecting a random split and scanning the head of that split? 

This seems to be testing the ability to descend the indices of compressed blocks then choosing random keys within there. Could we not remove the complexity here by simply scanning the beginning of the range? That could also make this a more general solution for different data models. @keith-turner would you disagree with this?

milleruntime:
Hey @tushardhadiwal thanks for the contribution!   At first glance, this code seems like it could be a useful scan timing measurement tool.  FYI Accumulo already has a general purpose availability monitor (named Monitor) in the [main repo](https://accumulo.apache.org/docs/2.x/administration/monitoring-metrics).  My only initial comment is that a different name, like "ScanProbe" or "ScanTimer", would be better to avoid confusion with the Monitor.

keith-turner:
Another path to consider is using the tserver ping function, but this does not do a scan.  Depends on what the desired behavior is.

https://accumulo.apache.org/docs/2.x/apidocs/org/apache/accumulo/core/client/admin/InstanceOperations.html#getTabletServers()

https://accumulo.apache.org/docs/2.x/apidocs/org/apache/accumulo/core/client/admin/InstanceOperations.html#ping(java.lang.String)

src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
```diff
@@ -0,0 +1,203 @@
...
+    String padded = splitStart.toString()
+        + "0000000000000000".substring(splitStart.toString().length());
+    Text minRow = new Text(genRow(Long.parseLong(padded.trim(), 16)));
+    return minRow;
+  }
+
+  public static Range pickRange(TableOperations tops, String table, Random r)
+      throws TableNotFoundException, AccumuloSecurityException, AccumuloException {
+
+    ArrayList<Text> splits = Lists.newArrayList(tops.listSplits(table));
```
phrocker: @keith-turner @tushardhadiwal Could you cache the splits? Or maybe we could investigate doing that into Accumulo with an observable of some type...but a low hanging fruit could be to cache the splits since they won't change very frequently with something like a loadercache in guava that expires at a certain point. The splits themselves, even if they were to change, would still give us the same information here -- so perhaps a long lived cache would be idea?

phrocker: As a side node I'm not sure hyped about caching in Accumulo for this function -- so having a cache here would be awesome.

tushardhadiwal: @phrocker Sure. I have pushed the commit to cache the splits.

src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
tushardhadiwal: @phrocker @keith-turner I agree. It would be nice to make this more general purpose & schema agnostic , but at the same time I would like to retain the functionality that alllows us to scan configurable number of rows. Can you throw some light on what would be the best way (function) to scan beginning of the range without comprising on ability to read configurable number of rows? A quick code blurb that elaborates more as a starter would be helpful.

tushardhadiwal: @phrocker @keith-turner Please take a look at the latest commit. Schema Agnostic , Scans configurable number of rows from beginning of the tablet. Caches tablet lookup information and reuses on later iterations.

tushardhadiwal:
> Another path to consider is using the tserver ping function, but this does not do a scan. Depends on what the desired behavior is.
> 
> https://accumulo.apache.org/docs/2.x/apidocs/org/apache/accumulo/core/client/admin/InstanceOperations.html#getTabletServers()
> 
> https://accumulo.apache.org/docs/2.x/apidocs/org/apache/accumulo/core/client/admin/InstanceOperations.html#ping(java.lang.String)

Desired behavior is to do actual scans.

tushardhadiwal:
> Hey @tushardhadiwal thanks for the contribution! At first glance, this code seems like it could be a useful scan timing measurement tool. FYI Accumulo already has a general purpose availability monitor (named Monitor) in the [main repo](https://accumulo.apache.org/docs/2.x/administration/monitoring-metrics). My only initial comment is that a different name, like "ScanProbe" or "ScanTimer", would be better to avoid confusion with the Monitor.

Hi, I understand your concern, I would change the name to something else once code is finalised and ready to merge, holding off on the name change for now until all comments are fully resolved.

src/main/java/org/apache/accumulo/testing/healthprobe/Monitor.java
```diff
@@ -0,0 +1,170 @@
...
+          }
+        } catch (Exception e) { 
+          log.error(String.format("Exception while scanning range %s. Check the state of Accumulo for errors.", range), e);
+        }
+      }
+    }
+  }
+
+  public static int consume(Iterable<Entry<Key, Value>> scanner, long numberOfRows) {
+    RowIterator rowIter = new RowIterator(scanner);
```
keith-turner: If you were interested in making this shorter, might be able to do the following for this method.

```java
  return Streams.stream(new RowIterator(scanner)).limit(numberOfRows).count();
```

README.md
```diff
@@ -238,6 +240,11 @@ test and produce json result files.
 There are some utilities for working with the json result files, run the `performance` script
 with no options to see them.
 
+## Availability Monitor 
+Monitor class aims at verifying availability of overall accumulo cluster by continually doing
+scans of random values across various tablet servers and capturing timing
+information related to how long such scans take. 
+
```
keith-turner: Could add an example command showing how to run.

tushardhadiwal: This will follow exact same pattern as cingest rwalk and other scripts. I have made the edits in the readme that show this like on Line 47 & Line 35

ctubbsii:
This appears to have been merged in c8ef7846b4426c5cb86632ade692b0ff8aa23977

phrocker:
> 
> 
> This appears to have been merged in [c8ef784](https://github.com/apache/accumulo-testing/commit/c8ef7846b4426c5cb86632ade692b0ff8aa23977)

Yes. I forgot the "This closes..." tag as I did from the command line. I had some mentions form long ago. Thanks for putting that in here @ctubbsii . I've become quite forgetful.

ctubbsii:
@phrocker You can also just do a squash and merge from the GitHub UI... that also gives you a chance to edit the commit message. This one wasn't well-formatted.

phrocker:
> 
> 
> @phrocker You can also just do a squash and merge from the GitHub UI... that also gives you a chance to edit the commit message. This one wasn't well-formatted.

ah yes. I wanted to sign the commit but after pushing I realized that wasn't necessary ( by way of seeing that other commits weren't signed )--  so yeah ... I should have used the UI ( though in reading about it I think that may be possible ). Either way, thanks!

PR #119, opened 2019-11-26
Add asynchronous input format and bulk key.

phrocker:
Currently validating some changes on this PR that I made as I submitted it. I am addressing some issues I found while running this on a cluster.


src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
```diff
@@ -60,25 +58,29 @@ public int run(String[] args) throws Exception {
...
 
     // remove bulk dir from args
     args = Arrays.asList(args).subList(1, 3).toArray(new String[2]);
 
     try (ContinuousEnv env = new ContinuousEnv(args)) {
+
+      /**
+       * Support various Input Formats
+       */
+      job.setInputFormatClass(env.getInputFormat());
```
phrocker: I made this change as I was submitting the PR so I will run some quick tests

src/main/java/org/apache/accumulo/testing/continuous/ThreadedContinousInputFormat.java
```diff
@@ -0,0 +1,237 @@
...
+
+import com.google.common.collect.Maps;
+
+/**
+ * Purpose and Justification: Defines an asynchronous input format that parallelizes the creation of
+ * the data. This aims to move the bar from a data generation + I/O ( spill ) problem to the right
+ * so that if we are able to optimize the spill path we ultimately minimize the impact of data
+ * generation.
+ *
+ * Note that this class leverages BulkKey versus Key. This was not merged into ContinousInputformat
```
phrocker: I was very cautious to keep the existing functionality. Adding threads to add threads doesn't always benefit so I wanted to isolate this functionality; however, I made a change while submitting this PR for ContinuousInputFormat to use BulkKey vs Key, so I will at the very least extend ContinuousInputFormat in ThreadedContinousInputFormat

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
```diff
@@ -0,0 +1,236 @@
...
+
+  @Override
+  public void write(DataOutput out) throws IOException {
+
+    key.getRow(row);
+    key.getColumnFamily(cf);
+    key.getColumnQualifier(cq);
+    key.getColumnVisibility(cv);
+
+    WritableUtils.writeVInt(out, row.getLength());
```
keith-turner: For the bulk ingest data, the row, family, and qualifier are all really small.  The length of each of these should fit in a single byte.  Could just always store the len in a single byte here instead of using writeVInt.  Seems like this  could simplify the code in KeyShortCircuitComparator.

Could have a little util method like 

```java
  void writeLen(OutputStream out, int len){
      Preconditions.checkArgument(len<128);
      out.write(len);
  }
```

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
```diff
@@ -0,0 +1,236 @@
...
+    WritableUtils.writeVInt(out, cf.getLength());
+    out.write(cf.getBytes(), 0, cf.getLength());
+
+    WritableUtils.writeVInt(out, cq.getLength());
+    out.write(cq.getBytes(), 0, cq.getLength());
+
+    WritableUtils.writeVInt(out, cv.getLength());
+    out.write(cv.getBytes(), 0, cv.getLength());
+
+    WritableUtils.writeVLong(out, key.getTimestamp());
```
keith-turner: Seems the current code is not setting the timestamp (not sure if that is  a good thing).  If its not being set, we could just not serialize it (possible with a sanity check).

Could also not serialize delete and instead do a sanity check to ensure its not a delete.

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
```diff
@@ -0,0 +1,236 @@
...
+      return EMPTY;
+    final byte[] data = new byte[size];
+    in.readFully(data, 0, data.length);
+    return data;
+  }
+
+  @Override
+  public void write(DataOutput out) throws IOException {
+
+    key.getRow(row);
```
keith-turner: This method calls Text.set() which copies into the array inside of text.  Could call getRowData(), which returns the array w/o copying.  Still allocating an object, but its probably no diff than allocating the Text object.

src/main/java/org/apache/accumulo/testing/continuous/AsyncRecordReader.java
```diff
@@ -0,0 +1,85 @@
...
+        // start up lazily
+        IntStream.range(0, threads).forEach(x -> {
+          Runnable generator = () -> {
+            CRC32 cksum = checksum ? new CRC32() : null;
+            while (generatedCount.longValue() < numNodes && running.get()) {
+              try {
+                final BulkKey key = genKey(cksum);
+                final Value value = new Value(createValue(uuid, key.getRowData().toArray(), cksum));
+                while (!queue.offer(Maps.immutableEntry(key, value), 1, TimeUnit.SECONDS)) {
+                  if (!running.get())
```
keith-turner: Does anything ever set this to false after its been set to true?

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
keith-turner: Could take this even further and assume the row, family, and qual are always the same len (with sanity checks) and not bother serializing the len.  This could make the compare in KeyShortCircuitComparator really simple and super fast (would not even need to worry about fields).

src/main/java/org/apache/accumulo/testing/continuous/AsyncRecordReader.java
phrocker: Let me double check. Thanks!

phrocker: I see a test failure so I'll check those too. Thanks for the review!

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
phrocker: Good ideas. Thanks!

phrocker:
I didn't create an issue, so allow me to provide justification for this PR. In reviewing running jobs I realized two things:

A) spills and sorts were spending an inordinate amount of time in the sort of Keys. To address this BulkKey will have a comparator that will reduce the deserialization time and memory.
B) Generating keys takes quite some time. Parallelization will allow us to create them asynchronously, while allowing the current code path to remain the same.

I ran these on a test system with very positive results. Happy to provide any benchmark results if desired. I think with Keith's ideas I'll likely be able to make some additional improvements.

keith-turner:
>  Happy to provide any benchmark results if desired.

I would like to see your results.

src/main/java/org/apache/accumulo/testing/continuous/BulkKey.java
keith-turner: @phrocker  curious if you ever got a chance to give this a try.

phrocker: @keith-turner I'm iteratively burning down my PRs. I will post updates to this on Friday with some testing. I have a sizable cluster where I can run this and post timing then. thanks!

phrocker: I was able to try it @keith-turner . Made a positive impact! Thanks! I'm actually adding some unit tests in the Key class, so I'll shoot a rebased and updated  PR Monday morning since it's gettin a bit late. This one was fun to test. 

FYI I took your suggestions of reducing the complexity of the key comparator and removing the timestamp/delete. I added some unit tests that function as regression tests in case any assumptions change in the testing framework. BulkKey was kind of general, but "TestKey" is very specific and focused on the specificities of size/type of keys.

keith-turner: Awesome.  Yeah a more specific name for the class would be good.  

> I added some unit tests that function as regression tests in case any assumptions change in the testing framework.

Nice, much better than runtime sanity checks.

phrocker: One thing I did not change was writing V int that's because it writes smaller sizes. One issue I ran into was spilling. If I use writeInt it writes 4 bytes, whereas writeVInt writes 1 byte for our sizes. While there are more branches it writes significantly less data to the context, resulting in overall greater throughput.

phrocker: I think that saved 12 bytes per key, which was enough to keep me from avoiding a second spill and drastically slowing my jobs down.

phrocker:
> > Happy to provide any benchmark results if desired.
> 
> I would like to see your results.

Let me re-gather them and I'll paste the timing and key output here. Thanks!

src/test/java/org/apache/accumulo/testing/continous/KeySerializationComparisonTest.java
```diff
@@ -0,0 +1,170 @@
...
+  @Test
+  public void validateComparison() {
+    TestKey.TestShortCircuitComparator comparator = new TestKey.TestShortCircuitComparator();
+    for (int i = 0; i < 1000; i++) {
+      byte[] a = WritableUtils.toByteArray(genTestKey(i, i));
+      int adder = random.nextInt(1001) + 1 + i;
+      byte[] b = WritableUtils.toByteArray(genTestKey(adder, adder));
+      int comparisonResult = comparator.compare(a, 0, a.length, b, 0, b.length);
+      Assert.assertTrue("Comparision should be negative, not " + comparisonResult,
+          comparisonResult < 0);
```
keith-turner: Would be nice to test when row same and cols differ.  Following may do that.
```java
 byte[] c = WritableUtils.toByteArray(genTestKey(i, adder));
 comparisonResult = comparator.compare(a, 0, a.length, c, 0, c.length);
￼ Assert.assertTrue("Comparision should be negative, not " + comparisonResult,
￼          comparisonResult < 0);
  // could compare c to b
```

src/test/java/org/apache/accumulo/testing/continous/KeySerializationComparisonTest.java
```diff
@@ -0,0 +1,170 @@
...
+  private static final long MEGABYTE = 10 * 1024L * 1024L;
+  Random random;
+
+  @Before
+  public void setup() {
+    random = new Random(System.currentTimeMillis());
+    visibilities = ContinuousIngest.parseVisibilities("IMATEST");
+  }
+
+  public static byte[] genCol(long cfInt) {
```
keith-turner: Could call `ContinuousIngest.genCol()` instead and avoid duplicating code.  Could statically import that function.

src/test/java/org/apache/accumulo/testing/continous/KeySerializationComparisonTest.java
```diff
@@ -0,0 +1,170 @@
...
+    return FastFormat.toZeroPaddedString(cfInt, 4, 16, EMPTY_BYTES);
+  }
+
+  public static long genLong(long min, long max, Random r) {
+    return ((r.nextLong() & 0x7fffffffffffffffL) % (max - min)) + min;
+  }
+
+  static byte[] genRow(long min, long max, Random r) {
```
keith-turner: Could statically import this from ContinuousIngest and avoid duplicating code.

src/test/java/org/apache/accumulo/testing/continous/BulkKey.java
```diff
@@ -0,0 +1,236 @@
...
+import org.apache.hadoop.io.Text;
+import org.apache.hadoop.io.WritableComparable;
+import org.apache.hadoop.io.WritableComparator;
+import org.apache.hadoop.io.WritableUtils;
+
+/**
+ * Supports immediate sorting via eager deser of the key object. This has the benefit of reducing
+ * the amount of deserialization that may occur when sorting keys in memory
+ */
+public class BulkKey implements WritableComparable<BulkKey> {
```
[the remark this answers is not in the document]
phrocker: yes it was only provided as comparison. I don't intend this to be in the final instance. I was just showing you some examples.

src/test/java/org/apache/accumulo/testing/continous/KeySerializationComparisonTest.java
phrocker: ha, yeah I was copying code around. Thanks for pointing this out!

src/main/java/org/apache/accumulo/testing/continuous/TestKey.java
```diff
@@ -0,0 +1,194 @@
...
+  static final byte[] EMPTY = {};
+
+  protected Text row = new Text();
+  protected Text cf = new Text();
+  protected Text cq = new Text();
+  protected Text cv = new Text();
+
+  public TestKey() {}
+
+  public TestKey(byte[] row, byte[] cf, byte[] cq, byte[] cv, long ts, boolean deleted) {
```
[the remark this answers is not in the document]
phrocker: Doh! I'm jumping around too much. Thanks for helping!

src/test/java/org/apache/accumulo/testing/continous/BulkKey.java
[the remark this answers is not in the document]
phrocker: It's used in the unit tests to provide a point of comparison and reference -- but was planning on removing it since you are the only one with any reference to what it is.

phrocker:
@keith-turner  I've run this a number of times in cases of three and 30 B keys generated. I tested increasing number of mappers the keys computed her mapper, etc. I'll focus on the three billion rows across 200 mappers for demonstration purposes only

Here is a sample of those test runs. I ran three sets of tests. One was the original setup without async threads or TestKey. The second was with async threads and BulkKey ( which I kept in the last commit for you to see that timing is roughly the same and we save a marginal amount of space ). The third set of runs were with async threads and BulkKey.

The async threads and specialized key comparator ( scenarios two and three ) won out in every case. The average runtime per mapper for all three scenarios were ( across 5 runs ). 
```
2.4 minutes 
1.45 minutes
1.5 minutes
```
The BulkKey and TestKey seem nearly identical; however, the amount of data written is dramatically different. For TestKey (scenario three) the aggregate output bytes were: 154,500,000,000. The other two resulted in 169,499,996,800 of key output. This also meant that when I doubled the number of keys, resulting in a single spill from TestKey, ( runtimes increased ) but the BulkKey and regular key objects actually resulted in a second spill that slowed down the job dramatically.

In my opinion async and the key comparator actually helped tremendously but your insight into optimizing TestKey to not output 12 additional bytes has a linear benefit on reducing spills and reducing transfer costs. 

In the case of the 30B runs the current ( regular job ) took about an hour and ten minutes. The async job with BulkKey took 77 minutes. The TestKey job took 70 minutes on a cluster that ran 200 map tasks. I certainly think the cluster can be optimized/improved -- but I ran these across a separate cluster of 100 VMs and while I obviously didn't have the same timing, the general trends were improved. 

I wonder if we could/should port BulkKey over to Accumulo Proper for map reduce jobs. I think there is definitely benefit in optimizing the comparison path ( and why I did the testing here). 

As far as fine grained profiling, Yourkit showed improvements in the write path but spills were the slowest part ( I/O ).

keith-turner:
@phrocker thanks for the writeup

> I wonder if we could/should port BulkKey over to Accumulo Proper for map reduce jobs.

I think that is a good idea. 

> In the case of the 30B runs the current ( regular job ) took about an hour and ten minutes. 

What test configuration was this?

ctubbsii:
This PR is stale, has conflicts, and hasn't been updated in over a year and a half. I'll close it now. If this still is a desired change, please rebase and re-submit.

phrocker:
> 
> 
> This PR is stale, has conflicts, and hasn't been updated in over a year and a half. I'll close it now. If this still is a desired change, please rebase and re-submit.

I forgot about this PR ( naturally -- that's what I do ha ). I want to make sure I don't misunderstand you. I'm interpreting this as submitting a new PR with the commits against main.

ctubbsii:
> I forgot about this PR ( naturally -- that's what I do ha ). I want to make sure I don't misunderstand you. I'm interpreting this as submitting a new PR with the commits against main. I think this PR is valuable but definitely wanted to make sure Keith didn't have any issues with it. Happy to resolve conflicts and submit a new one or to push back into this PR and re-open.

Either way works. I have no preference for a new PR or re-opening and rebasing in-place. Whatever is more convenient.

PR #121, opened 2019-11-27
Fix #120 Support non default absolute hdfs paths in cingest bulk

### LINKED ISSUE #120: Support non default HDFS path as a parameter for continuous bulk ingest
Support non default HDFS (in case of multiple volumes) path as a parameter for continuous bulk ingest. For e.g
`bin/cingest bulk abfs://data@azaccumulo.dfs.core.windows.net/azbulk-multi` 

In this case `abfs://data@azaccumulo.dfs.core.windows.net/`  HDFS filesystem is not the default HDFS and has been added as a additional volume to accumulo
The default HDFS fileystem configured is `hdfs://accucluster`

`Exception in thread "main" java.lang.IllegalArgumentException: Wrong FS: abfs://data@azaccumulo.dfs.core.windows.net/azbulk-multi, expected: hdfs://accucluster:8020`

mandarinamdar:
Support non default and/or absolute hdfs paths in continuous bulk ingestion

* Accept non default/absolute hdfs paths too as a parameter to cingest bulk
* Without this change, only defaultFS works
* Without this change, it gives error for any parameter other than relative path
* This feature support is critical, especially in case of multiple hdfs volumes

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
@@ -19,6 +19,7 @@ package org.apache.accumulo.testing.continuous;
 
 import java.io.BufferedOutputStream;
 import java.io.PrintStream;
+import java.net.URI;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collection;
@@ -50,13 +51,14 @@ public class BulkIngest extends Configured implements Tool {
   @Override
   public int run(String[] args) throws Exception {
     String ingestInstanceId = UUID.randomUUID().toString();
+    String bulkDir = args[0];
 
     Job job = Job.getInstance(getConf());
     job.setJobName("BulkIngest_" + ingestInstanceId);
     job.setJarByClass(BulkIngest.class);
     // very important to prevent guava conflicts
     job.getConfiguration().set("mapreduce.job.classloader", "true");
-    FileSystem fs = FileSystem.get(job.getConfiguration());
+    FileSystem fs = FileSystem.get(URI.create(bulkDir), job.getConfiguration());
 
     log.info(String.format("UUID %d %s", System.currentTimeMillis(), ingestInstanceId));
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
```diff
@@ -50,13 +51,14 @@
...
     String ingestInstanceId = UUID.randomUUID().toString();
+    String bulkPath = args[0];
 
     Job job = Job.getInstance(getConf());
     job.setJobName("BulkIngest_" + ingestInstanceId);
     job.setJarByClass(BulkIngest.class);
     // very important to prevent guava conflicts
     job.getConfiguration().set("mapreduce.job.classloader", "true");
-    FileSystem fs = FileSystem.get(job.getConfiguration());
+    FileSystem fs = FileSystem.get(URI.create(bulkPath), job.getConfiguration());
```
keith-turner: Do you know if paths w/o a scheme like `/tmp/bt/1` still work?

src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
```diff
@@ -76,7 +78,8 @@ public int run(String[] args) throws Exception {
 
       // output RFiles for the import
       job.setOutputFormatClass(AccumuloFileOutputFormat.class);
-      AccumuloFileOutputFormat.configure().outputPath(new Path(bulkDir + "/files")).store(job);
+      AccumuloFileOutputFormat.configure().outputPath(new Path(fs.getUri() + bulkDir + "/files"))
```
keith-turner: With the following, may not need `bulkDir`.  I think the `makeQualified` fuction only prefixes the fs URI if its not there.

```suggestion
      AccumuloFileOutputFormat.configure().outputPath(fs.makeQualified(new Path(bulkPath + "/files")))
```

src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
mandarinamdar: @keith-turner , thanks for reviewing. Yes, it works w/o scheme, with (default or otherwise) scheme as well

src/main/java/org/apache/accumulo/testing/continuous/BulkIngest.java
mandarinamdar: I will modify and check

mandarinamdar: I have changed the code to use fs.makeQualified consistently and reuse existing bulkDir variable avoiding new variable bulkPath

Issue #122, opened 2020-01-24
Missing license headers


billierinaldi:
There are several java files missing license headers in the accumulo-testing repo. We should add headers as needed and [add the apache-rat-plugin to the pom.xml](https://creadur.apache.org/rat/apache-rat-plugin/examples/verify.html) so that CI will catch this problem when running mvn verify.

jmark99:
Closed by #124

PR #124, opened 2020-09-22
Missing license headers #122

jmark99:
* Added missing license headers to 20+ files.
* Added the apache-rat plugin to the pom file.
* Moved the splits data files in test/bench/lib into a data
subdirectory.
* Modified TableSplitsBenchmark.py to account for the movement of the
splits file into the data directory.
* Added exclusions for the data directory and the exclude_classes.txt
file.


.gitignore
```diff
@@ -5,3 +5,5 @@
 /*.iml
 /.idea
 /logs/
+/test/bench/lib/*.pyc
+/test/bench/cloudstone*/*.pyc
```
ctubbsii, line 9: This might be better as:
```suggestion
*.pyc
```

ctubbsii (approved):

PR #139, opened 2021-04-14
Create ZKDroppingIT

milleruntime:
Created this in an attempt to reproduce https://github.com/apache/accumulo/issues/1994

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,16 @@
       <version>${accumulo.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.apache.accumulo</groupId>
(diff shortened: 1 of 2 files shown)


milleruntime:
This test will currently hang after the ZK node is triggered and eventually time out. In a way this test is successful at what it intends to do. Once the ZK node is triggered, the BadZK will trigger exceptions, putting the tserver in a half alive state. It will get stuck at flushing any data but it will still maintain a lock in ZK so the Manager won't kill it.

milleruntime:
Even if I get this working as intended, I don't think it will be possible to create a reliable test from it.

PR #142, opened 2021-06-29
Allow kerberized run

BukrosSzabolcs:
Adding login functionality to client creation and passing the delegation
token when yarn is used to allow running tests on a kerberized cluster.
(security tests excluded)

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/KerberosHelper.java
+++ b/src/main/java/org/apache/accumulo/testing/KerberosHelper.java
@@ -1,0 +1,67 @@
+package org.apache.accumulo.testing;
+
+import java.io.IOException;
+import java.util.Properties;
+
+import org.apache.accumulo.core.client.Accumulo;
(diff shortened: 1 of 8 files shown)


dlmarion:
This looks ok to me. I tried to find a class in the `accumulo` repo that could be used in place of KerberosHelper, but I'm not sure one exists. My only question is "should all of the classes in `accumulo-testing` be set up in the same manner, to enable them to run in an environment running Kerberos?". I think the answer is yes, but it could be done after this change, maybe a good first issue ticket.

dlmarion (approved):

ctubbsii:
Merging based on Dave's previous review. I also skimmed it and it seemed fine.

ctubbsii:
Thanks @BukrosSzabolcs !

BukrosSzabolcs:
Thanks @dlmarion for the review and @ctubbsii for the merge!

PR #143, opened 2021-07-10
Update guava version to 30.1.1 (matches accumulo project)

EdColeman:


ctubbsii:
The test that failed was a replication-related test. Replication-related tests were disabled in the main repo because of their unreliability, and lack of active maintainer working on that feature. Perhaps they should be disabled here as well?

Manno15:
I believe the failing IT was fixed in #147. So this should be able to pass now if updated with main.

EdColeman:
Thank @Manno15 - I re-based and pushed to see if it will pass cleanly.

Manno15 (approved):

PR #146, opened 2021-08-02
Migrate accumulo-testing to log4j2.

### LINKED ISSUE #141: Migrate to log4j2
Follow on from issue #130 / PR #140

This testing repository should migrate to log4j2, and any configured console logging should be configured to use STDERR instead of STDOUT in the log4j2 configuration files, so that console output won't interfere with output from executed commands used for scripts (see #130).

AlbertWhitlock:
Migrate accumulo-testing to use log4j2. Fixes #141.


pom.xml
```diff
@@ -374,6 +439,8 @@
                   <transformers>
                     <!-- Hadoop uses service loader to find filesystem impls, without this may not find them -->
                     <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
+                    <!-- New transformer. Stops "unrecognized format specifier" error. -->
```
ctubbsii, line 442: Let's add a comment with a link to `https://github.com/edwgiz/maven-shaded-log4j-transformer` in case anybody has questions about what this does. The README there should answer any additional questions.

pom.xml
```diff
@@ -94,15 +106,51 @@
...
     </dependency>
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
-      <version>${slf4j.version}</version>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-bom</artifactId>
+      <version>2.14.1</version>
+      <type>pom</type>
+      <scope>runtime</scope>
```
ctubbsii, line 153: BOM files should be used with the import scope, and declared in the `<dependencyManagement />` section. The version here could use the `${log4j2.version}` property instead of specifying it explicitly here.

pom.xml
```diff
@@ -94,15 +106,51 @@
       <artifactId>hadoop-client-runtime</artifactId>
       <version>${hadoop.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+      <version>${log4j2.version}</version>
+    </dependency>
```
ctubbsii, line 113: Specifying the BOM in the dependencyManagement section eliminates the need to specify any versions for any of these log4j2 jars in this section.

pom.xml
```diff
@@ -94,15 +106,51 @@
...
+      <artifactId>log4j-slf4j-impl</artifactId>
+      <version>${log4j2.version}</version>
+    </dependency>
+    <!--
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-1.2-api</artifactId>
+      <version>${log4j2.version}</version>
+    </dependency>
+    -->
```
ctubbsii, line 142: Please remove any unused commented out entries.

pom.xml
```diff
@@ -48,6 +50,11 @@
     </dependencies>
   </dependencyManagement>
   <dependencies>
+    <dependency>
+      <groupId>com.github.edwgiz</groupId>
+      <artifactId>maven-shade-plugin.log4j2-cachefile-transformer</artifactId>
+      <version>2.1</version>
+    </dependency>
```
ctubbsii, line 57: This should not be specified as a dependency of this project, as it is only a dependency of the shade plugin, in order to add the transformer to that.

pom.xml
```diff
@@ -29,12 +29,14 @@
   <description>Testing tools for Apache Accumulo</description>
   <properties>
     <accumulo.version>2.1.0-SNAPSHOT</accumulo.version>
+    <disruptor.version>3.4.4</disruptor.version>
     <eclipseFormatterStyle>${project.basedir}/contrib/Eclipse-Accumulo-Codestyle.xml</eclipseFormatterStyle>
     <hadoop.version>3.2.0</hadoop.version>
+    <log4j2.version>2.14.1</log4j2.version>
```
ctubbsii, line 35: Not sure if it's useful to have a property for this, if you use the BOM, since the BOM should eliminate the need to use this property more than one time. However, we may still want to be able to set the value with `-Dlog4j2.version=` when building to override this, in which case, the property would still need to be here, even if it's only ever used once (for the BOM).

conf/log4j2.properties
```diff
@@ -0,0 +1,57 @@
...
+appender.console.target = SYSTEM_ERR
+appender.console.layout.type = PatternLayout
+# appender.console.layout.pattern = %style{%d{ISO8601}}{dim,cyan} %style{[}{red}%style{%-8c{2}}{dim,blue}%style{]}{red} %highlight{%-5p}%style{:}{red} %m%n
+appender.console.layout.pattern = %d{ISO8601} [%c{3}] %-5p: %m%n
+
+logger.accumulo.name = org.apache.accumulo
+logger.accumulo.level = WARN
+logger.accumulo.testing = DEBUG
+# logger.org.apache.accumulo=WARN
+#logger.org.apache.accumulo.testing=DEBUG
```
ctubbsii, line 36: Should remove any of these commented out lines that aren't used.

conf/log4j2.properties
```diff
@@ -0,0 +1,57 @@
...
+appender.console.type = Console
+appender.console.name = STDERR
+appender.console.target = SYSTEM_ERR
+appender.console.layout.type = PatternLayout
+# appender.console.layout.pattern = %style{%d{ISO8601}}{dim,cyan} %style{[}{red}%style{%-8c{2}}{dim,blue}%style{]}{red} %highlight{%-5p}%style{:}{red} %m%n
+appender.console.layout.pattern = %d{ISO8601} [%c{3}] %-5p: %m%n
+
+logger.accumulo.name = org.apache.accumulo
+logger.accumulo.level = WARN
+logger.accumulo.testing = DEBUG
```
ctubbsii, line 34: This isn't quite right. This should be:

```suggestion
logger.accumulotesting.name = org.apache.accumulo.testing
logger.accumulotesting.level = DEBUG
```

The segment after `logger.` is a way of creating a named reference to a logger. Once you have a reference, you need to tell log4j2 what logger name pattern it refers to with `.name = ` and what minimum level to accept for that logger reference with `.level = `.

So, in general:

```ini
logger.refA.name = some.prefix.pattern.here
logger.refA.level = DEBUG
logger.refB.name = some.other.prefix.here
logger.refB.level = WARN
```

In log4j1, you would have put the pattern in the logger reference itself, but you no longer do that in log4j2, since the logger reference is created explicitly with `logger.refName.name = pattern`. So, `logger.accumulo.testing = DEBUG` doesn't do anything, since `.testing` isn't valid.

This error is repeated a few times below. Hopefully this explanation helps.

ctubbsii (changes requested):
I was a little confused by the log4j2 plugin cache transformer stuff, as I never had to deal with that before, but realized it's needed specifically because we're shading in the log4j code itself into the shaded jar.

conf/log4j2.properties
```diff
@@ -0,0 +1,51 @@
...
+name = AccumuloTestingDefaultLoggingProperties
+
+appender.console.type = Console
+appender.console.name = STDERR
+appender.console.target = SYSTEM_ERR
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{ISO8601} [%c{3}] %-5p: %m%n
+
+logger.accumulotesting.name = org.apache.accumulo.testing
+logger.accumulotesting.level = DEBUG
```
ctubbsii, line 32: The old file specified both `org.apache.accumulo` and `org.apache.accumulo.testing`. We probably need both, with the former set to WARN, as in the old file.

ctubbsii (changes requested):
Looking good. Nearly done.

conf/log4j2.properties
```diff
@@ -0,0 +1,54 @@
...
+
+logger.hadooptest.name = org.apache.hadoop
+logger.hadooptest.level = WARN
+logger.hadoopmapred.name = org.apache.hadoop.mapreduce
+logger.hadoopmapred.level = ERROR
+logger.hadooputil.name = org.apache.hadoop.util
+logger.hadooputil.level = ERROR
+
+logger.zookeepertest.name = org.apache.zookeeper
+logger.zookeepertest.level = WARN
```
Manno15, line 45: The old file had zookeeper as `ERROR` instead of `WARN`. Not sure how big of a deal that is though.

Manno15:
I am not quite getting some log messages I expect to get with continuous ingest. I suspect the appender is only sending errors to console and not info messages. I will take a look further to make sure it's not just on my end.

conf/log4j2.properties
```diff
@@ -0,0 +1,54 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+## Log4j2 file that configures logging for processes other than Accumulo services
```
Manno15, line 20: This line is taken from log4j2.properties from the main accumulo repo. It doesn't seem accurate here since there is only one properties file for every accumulo-testing operation.

src/main/java/org/apache/accumulo/testing/randomwalk/Framework.java
```diff
@@ -25,6 +25,7 @@
 public class Framework {
 
   private static final Logger log = LoggerFactory.getLogger(Framework.class);
+
```
Manno15, line 28:
```suggestion
```

Manno15:
Your recent changes look good. I am now getting the logs I expect to see.

ctubbsii (approved):
LGTM. Will merge based on @Manno15 's observation that the log output works as expected. Did not manually test, but did thoroughly examine the changes, and they look right.

PR #147, opened 2021-08-09
Replace deprecated prop inside ReplicationRandomWalkIT

Manno15:
This test was timing out and seemingly due to its use of the deprecated property `TSERV_WALOG_MAX_SIZE` (now `TSERV_WAL_MAX_SIZE`).


Manno15:
We should probably look for other uses of the deprecated property inside accumulo-testing and resolve those issues similar to changes made in PR https://github.com/apache/accumulo/pull/2171/

DomGarguilo (approved):
I was able to reproduce the test failure and verified that this change makes it pass.

PR #149, opened 2021-08-17
 Update CI to use Compaction Planner

### LINKED ISSUE #145: Update CI to use Compaction Planner
Continuous Ingest is still using `BasicCompactionStrategy`, which has been deprecated in favor of the new compaction code. This needs to be replaced with a CompactionPlanner, most likely `DefaultCompactionPlanner` will work fine. This may just be a configuration change.

This would be great to do for the 2.1 release testing.

DomGarguilo:
Fixes #145 

Changes were added to allow the compaction planner configuration to be changed via the properties file.

The current settings were taken from [an example](https://accumulo.apache.org/blog/2021/07/08/external-compactions.html#configuration-1) and may not be optimal for this specific test. Not sure what is best for this case and recommendations are welcome.

### DIFF
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -42,7 +42,10 @@ test.ci.common.auths=
 # Accumulo tserver properties to set when creating a table
 test.ci.common.accumulo.server.props=\
 tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
-tserver.compaction.major.service.cs1.planner.opts.executors=[{"name":"small","type":"internal","maxSize":"32M","numThreads":4},{"name":"medium","type":"internal","maxSize":"128M","numThreads":2},{"name":"large","type":"external","queue":"DCQ1"}]
+tserver.compaction.major.service.cs1.planner.opts.executors=\
+[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
+{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
+{"name":"large","type":"internal","numThreads":2}]
   
 # Accumulo table properties to set when creating table
 test.ci.common.accumulo.table.props=\
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -39,12 +39,15 @@ test.ci.common.accumulo.num.tablets=20
 # Optional authorizations that if specified will be randomly selected by scanners and walkers
 # Format: a,b|a,b,c|c
 test.ci.common.auths=
+# Accumulo tserver properties to set when creating a table
+test.ci.common.accumulo.server.props=\
+tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
+tserver.compaction.major.service.cs1.planner.opts.executors=[{"name":"small","type":"internal","maxSize":"32M","numThreads":4},{"name":"medium","type":"internal","maxSize":"128M","numThreads":2},{"name":"large","type":"external","queue":"DCQ1"}]
+  
 # Accumulo table properties to set when creating table
 test.ci.common.accumulo.table.props=\
-table.majc.compaction.strategy=org.apache.accumulo.tserver.compaction.strategies.BasicCompactionStrategy \
-table.majc.compaction.strategy.opts.filter.size=250M \
-table.majc.compaction.strategy.opts.large.compress.threshold=100M \
-table.majc.compaction.strategy.opts.large.compress.type=gz
+table.compaction.dispatcher=org.apache.accumulo.core.spi.compaction.SimpleCompactionDispatcher \
+table.compaction.dispatcher.opts.service=cs1
 
 # Ingest
 # ------
--- a/src/main/java/org/apache/accumulo/testing/continuous/CreateTable.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/CreateTable.java
@@ -59,13 +59,14 @@ public class CreateTable {
 
       // retrieve and set tserver props
       Map<String,String> props = getProps(env, TestProps.CI_COMMON_ACCUMULO_SERVER_PROPS);
-      props.forEach((k, v) -> {
-        try {
-          client.instanceOperations().setProperty(k, v);
-        } catch (AccumuloException | AccumuloSecurityException e) {
-          e.printStackTrace();
+      try {
+        for (Map.Entry<String,String> entry : props.entrySet()) {
+          client.instanceOperations().setProperty(entry.getKey(), entry.getValue());
         }
-      });
+      } catch (AccumuloException | AccumuloSecurityException e) {
+        log.error("Failed to set tserver props");
+        throw new Exception(e);
+      }
 
       SortedSet<Text> splits = new TreeSet<>();
       final int numSplits = numTablets - 1;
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/accumulo/testing/continuous/CreateTable.java
```diff
@@ -55,9 +57,19 @@ public static void main(String[] args) throws Exception {
         System.exit(-1);
       }
 
+      // retrieve and set tserver props
+      Map<String,String> props = getProps(env, TestProps.CI_COMMON_SERVER_PROPS);
+      props.forEach((k, v) -> {
+        try {
+          client.instanceOperations().setProperty(k, v);
+        } catch (AccumuloException | AccumuloSecurityException e) {
+          e.printStackTrace();
```
DomGarguilo, line 66: Also not too sure how this case could be best handled.

ctubbsii: 1. Instead of using `.forEach`, you can just use a regular for loop over the `props.entrySet()`, so you can throw the exceptions out of the loop.
2. Or, you could log the error and stack trace using the logger, and call System.exit, like in other scenarios (not a fan of that one).
3. Or just wrap and rethrow as a RuntimeException whose cause is `e`.

Of the options, I think the first is best. `.forEach` is convenient in some cases, but it doesn't play nice with checked exceptions, so it's often just cleaner to use traditional loops.

milleruntime (approved):
I haven't ran CI with the new compaction stuff but this looks like a good start. We can always tweak settings later after doing release testing

Manno15:
In my testing I did run into several similar errors: 

 ```
returned non-existant compaction service cs1 for compaction type SELECTOR. 
Check the table compaction dispatcher configuration. Attempting to fall back to default service.
```

It appears the compaction service isn't being created properly.

**Comment by DomGarguilo** [comment]
> It appears the compaction service isn't being created properly.

Good catch, I was only verifying using `config` on the table created by `cingest createtable` which shows that the new properties are set which shows:


```
site       | tserver.compaction.major.service.cs1.planner ..................... | 
system     |    @override ..................................................... | org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
site       | tserver.compaction.major.service.cs1.planner.opts.executors ...... | 
system     |    @override ..................................................... | [{"name":"small","type":"internal","maxSize":"32M","numThreads":4},{"name":"medium","type":"internal","maxSize":"128M","numThreads":2},{"name":"large","type":"external","queue":"DCQ1"}]
```

My guess is the compaction service `cs1` is not being properly defined (where i tried `client.instanceOperations().setProperty()`) so when `cs1` is set to the table, it errors out.

**Comment by Manno15** [comment]
After a few rebuilds, it does seem the service is being created properly but now I am running into 

```
java.lang.IllegalArgumentException: Compacting not in set of all files
 [[A00000cj.rf, 31769681 814240], [F00000cy.rf, 2463619 65081]] [[F00000c7.rf, 2164845 57654], 
[C000008s.rf, 9949180 253514], [C00000c0.rf, 9967543 253962], [A0000069.rf, 9284906 249110]]
```
Which seems to be related to the properties being set. I ran [the example](https://accumulo.apache.org/blog/2021/07/08/external-compactions.html#configuration-1) manually and it also errored out with the same error as above.

**Comment by milleruntime** [comment]
@Manno15 Are the issues you seeing with the configuration? Or do you think there is a bug in the Compaction Planner?

**Comment by dlmarion** [comment]
@DomGarguilo  - did you see https://github.com/apache/accumulo-website/commit/b701adfa371d375f82813b2834af9e63894fd789?  I updated the docs to add examples.

**Comment by dlmarion** [comment]
There are also examples at https://accumulo.apache.org/blog/2021/07/08/external-compactions.html

**Comment by DomGarguilo** [comment]
> @DomGarguilo - did you see [apache/accumulo-website@b701adf](https://github.com/apache/accumulo-website/commit/b701adfa371d375f82813b2834af9e63894fd789)? I updated the docs to add examples.

I don't think I saw this but...

> There are also examples at https://accumulo.apache.org/blog/2021/07/08/external-compactions.html

I did reference these examples when creating these changes. @dlmarion

**Comment by Manno15** [comment]
> @Manno15 Are the issues you seeing with the configuration? Or do you think there is a bug in the Compaction Planner?

@milleruntime I haven't looked into it close enough to tell but I will investigate further and compare with different examples and values.

**Comment by Manno15** [comment]
@milleruntime I am leaning on this being a bug with either compaction Planner or another aspect of the compaction changes. I was able to reproduce the errors by creating a table using the shell and not changing anything else about the cluster. The error will eventually occur after ingesting for several minutes. It is possible the default compaction planner properties is the issue but that needs to be investigated further. I will create an issue in accumulo to track this as the issue seems unrelated to the changes made in this PR.

File: conf/accumulo-testing.properties [hunk 2]
```diff
@@ -39,12 +39,15 @@ test.ci.common.accumulo.num.tablets=20
 # Optional authorizations that if specified will be randomly selected by scanners and walkers
 # Format: a,b|a,b,c|c
 test.ci.common.auths=
+# Accumulo tserver properties to set when creating a table
+test.ci.common.accumulo.server.props=\
+tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
+tserver.compaction.major.service.cs1.planner.opts.executors=[{"name":"small","type":"internal","maxSize":"32M","numThreads":4},{"name":"medium","type":"internal","maxSize":"128M","numThreads":2},{"name":"large","type":"external","queue":"DCQ1"}]
```
Manno15, line 45: After looking into this more, we should not include an external queue as the default case unless we also include starting the other compaction services (compactor and compaction coordinator) by default. I think the best case is to have the configuration only be internal opts and then the user can decide on if they want external compactions and then start the other services.

dlmarion: agreed, if you set external but don't start the external services then some compactions will not occur.

Manno15: It would probably be beneficial to also leave a comment with either an example of opts with an external queue or a link to the website example.

DomGarguilo: addressed this in a51ef07

src/main/java/org/apache/accumulo/testing/continuous/CreateTable.java
DomGarguilo: addressed this in a51ef07 @ctubbsii

Manno15 (approved):
Changes look good. The configuration could possibly be fine-tuned but that can be adjusted later. I think this can be merged now as the bug seems unrelated to these changes.

src/main/java/org/apache/accumulo/testing/continuous/CreateTable.java
```diff
@@ -55,9 +57,20 @@ public static void main(String[] args) throws Exception {
         System.exit(-1);
       }
 
+      // retrieve and set tserver props
+      Map<String,String> props = getProps(env, TestProps.CI_COMMON_ACCUMULO_SERVER_PROPS);
+      try {
+        for (Map.Entry<String,String> entry : props.entrySet()) {
+          client.instanceOperations().setProperty(entry.getKey(), entry.getValue());
+        }
```
DomGarguilo, line 65: Looking back at this, I'm wondering if an `instanceOperations().setProperties()` method would be useful. Haven't looked into how often it could be used but it would be clean up cases like this where multiple properties need to be set.

Manno15:
> I will probably wait to see if accumulo-2247 is merged before merging this PR so I can make use of the new method here.

@DomGarguilo I think this should go ahead and be merged now since [accumulo-2247 ](https://github.com/apache/accumulo/pull/2247)  is also waiting on a different PR. These changes are beneficial to have in accumulo-testing as is and any improvements can be added in later once the other PR is resolved.

PR #150, opened 2021-09-08
Refactor out Non API code

### LINKED ISSUE #53: Refactor out Non API code
With the import-control checkstyle, it is now easier to detect use of non-API code used. The following exceptions should be refactored and removed from the config file:
https://github.com/apache/accumulo-testing/blob/master/contrib/import-control.xml#L35

DomGarguilo:
Fixes #53 

Refactored out non-api code. 

In most cases the imports were just being used to grab a String e.g. a class name. For those cases I just hardcoded a `final String` to use in their place.


DomGarguilo:
I am still looking into how FastFormat might be refactored out. I know that String.format is a lot slower so wanted to get some feedback on if that is acceptable to use in this case or if something else should be done such as creating a method to reproduce the functionality of FastFormat.

src/main/java/org/apache/accumulo/testing/merkle/MerkleTree.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Arrays;
 import java.util.List;
 
-import org.apache.accumulo.core.util.Pair;
+import org.apache.commons.math3.util.Pair;
```
milleruntime, line 24: Nice find!

src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
```diff
@@ -36,6 +35,7 @@
   private static final int ONE_SECOND = 1000;
   private static final String TABLE_NAME = "highSplitCreation";
   private static final String METADATA_TABLE_SPLITS = "123456789abcde";
+  private static final String METADATA_TABLE_NAME = "accumulo.metadata";
```
milleruntime, line 38: I think this is fine but we could probably put the names in a shared place of constants and reuse them.

milleruntime (approved):
Looks good. Thanks for looking into this! As for the FastFormat class, that is a quite useful utility. I am fine with leaving it as an exception. The easiest solution would be to just copy the class into testing since I doubt it will be changing any time soon.

src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
DomGarguilo: > I think this is fine but we could probably put the names in a shared place of constants and reuse them.

Good idea. Do you know of what a good place for these might be? @milleruntime

Manno15: TestProps.java seems like a good location.

DomGarguilo:
> The easiest solution would be to just copy the class into testing since I doubt it will be changing any time soon.

I went ahead and did that. All of the exceptions should be accounted for now.

src/main/java/org/apache/accumulo/testing/performance/tests/HighSplitCreationPT.java
DomGarguilo: Addressed in c8b6cf4

contrib/import-control.xml
```diff
@@ -37,18 +37,6 @@
...
     <allow class="org.apache.accumulo.core.util.threads.ThreadPools"/>
 
-    <!-- TODO refactor code to remove the following exceptions -->
-    <allow class="org.apache.accumulo.core.metadata.MetadataTable"/>
-    <allow class="org.apache.accumulo.core.replication.ReplicationTable"/>
-    <allow class="org.apache.accumulo.core.spi.scan.HintScanPrioritizer"/>
-    <allow class="org.apache.accumulo.core.clientImpl.TabletServerBatchWriter"/>
-    <allow class="org.apache.accumulo.core.util.FastFormat"/>
-    <allow class="org.apache.accumulo.core.util.Pair"/>
-    <allow class="org.apache.accumulo.core.trace.Trace"/>
```
DomGarguilo, line 47: As is, there are commented out usages of Trace in [VerifyIngest.java](https://github.com/apache/accumulo-testing/blob/fbaf16dc2c065eaab690d8d7abd75750d5d0425c/src/main/java/org/apache/accumulo/testing/ingest/VerifyIngest.java#L65). I'm thinking that either the comments should be removed along with the `<allow/>` tag or, both should be kept. For now I'll remove both and change it back if it seems like it should be.

milleruntime: My guess is that was commented out because either it wasn't working or the Trace code was in flux (it still is). Since there is an option to use the Trace in VerifyIngest, we could throw an exception stating it is currently not supported.

DomGarguilo: Something like 96ff1ac? @milleruntime

src/main/java/org/apache/accumulo/testing/ingest/VerifyIngest.java
```diff
@@ -59,6 +59,9 @@ public static void main(String[] args) throws Exception {
     Opts opts = new Opts();
     opts.parseArgs(VerifyIngest.class.getName(), args);
     try (AccumuloClient client = Accumulo.newClient().from(opts.getClientProps()).build()) {
+      if (opts.trace) {
+        throw new RuntimeException("Trace is enabled but not currently supported");
```
EdColeman, line 63: could use UnsupportedOperationException

DomGarguilo: Good idea. Addressed in 02248ad

milleruntime (approved):
Nice improvements!

PR #151, opened 2021-09-09
Replace use of the deprecated `newInstance()` method

DomGarguilo:
While working on #150 I noticed the following while building:
```
[WARNING] /home/dgarguilo/github/accumulo-testing/src/main/java/org/apache/accumulo/testing/scalability/Run.java:[83,9] newInstance() in java.lang.Class has been deprecated
```

It looks like `newInstance()` has been deprecated and replaced by `getDeclaredConstructor().newInstance()`.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/scalability/Run.java
+++ b/src/main/java/org/apache/accumulo/testing/scalability/Run.java
@@ -80,7 +80,7 @@ public class Run {
 
     ScaleTest test = (ScaleTest) Class
         .forName(String.format("org.apache.accumulo.test.scalability.%s", opts.testId))
-        .newInstance();
+        .getDeclaredConstructor().newInstance();
 
     test.init(scaleProps, testProps, opts.numTabletServers);
 


Manno15 (approved):

milleruntime:
The Scalability test is pretty old and I don't know of anyone that has run the test in a while. It might be worth it as a follow on to run it and update it. It may be still be a useful test.

DomGarguilo:
> The Scalability test is pretty old and I don't know of anyone that has run the test in a while. It might be worth it as a follow on to run it and update it. It may be still be a useful test.

I'll go ahead and merge this for now then take a closer look at the scalability test. Was there anything in particular you had in mind in terms of updating it?

milleruntime:
> > The Scalability test is pretty old and I don't know of anyone that has run the test in a while. It might be worth it as a follow on to run it and update it. It may be still be a useful test.
> 
> I'll go ahead and merge this for now then take a closer look at the scalability test. Was there anything in particular you had in mind in terms of updating it?

Mostly just test to make sure it still works. I took a quick look at the classes and one thing we could do is make a helper script in `<testing-home>/bin` like we do with the other tests. That would make it easier to run at least.

PR #152, opened 2021-09-10
Added check for existing output directory in /bin/cingest verify

### LINKED ISSUE #92: bin/cingest verify fails with 'Output directory hdfs://localhost:8020/tmp/ci-verify already exists'
If I've already run a continuous ingest verify and I want to run it again, I get the following error:

```java
Exception in thread "main" org.apache.hadoop.mapred.FileAlreadyExistsException: Output directory hdfs://localhost:8020/tmp/ci-verify already exists
	at org.apache.hadoop.mapreduce.lib.output.FileOutputFormat.checkOutputSpecs(FileOutputFormat.java:164)
	at org.apache.hadoop.mapreduce.JobSubmitter.checkSpecs(JobSubmitter.java:277)
	at org.apache.hadoop.mapreduce.JobSubmitter.submitJobInternal(JobSubmitter.java:143)
	at org.apache.hadoop.mapreduce.Job$11.run(Job.java:1570)
	at org.apache.hadoop.mapreduce.Job$11.run(Job.java:1567)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1730)
	at org.apache.hadoop.mapreduce.Job.submit(Job.java:1567)
	at org.apache.hadoop.mapreduce.Job.waitForCompletion(Job.java:1588)
	at org.apache.accumulo.testing.continuous.ContinuousVerify.run(ContinuousVerify.java:193)
	at org.apache.hadoop.util.ToolRunner.run(ToolRunner.java:76)
	at org.apache.accumulo.testing.continuous.ContinuousVerify.main(ContinuousVerify.java:204)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.apache.hadoop.util.RunJar.run(RunJar.java:323)
	at org.apache.hadoop.util.RunJar.main(RunJar.java:236)
```

It'd be better if the script noticed that a verify job had already run, and asked to clean up this directory for me, so I didn't have to manually execute: `hdfs dfs -rm -r /tmp/ci-verify`

DomGarguilo:
Fixes #92 

* ~~Added a check to see if the output directory exists. If so, the user is prompted to see if they want the directory to be removed and the verify job rerun.~~
* EDIT: Create a unique output directory to store the output of each run of `/bin/cingest verify`
* Made applicable variables final. Unrelated to the issue addressed but thought I would improve the file while I was already making changes

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
@@ -52,9 +52,11 @@ import org.slf4j.LoggerFactory;
 public class ContinuousVerify extends Configured implements Tool {
   public static final VLongWritable DEF = new VLongWritable(-1);
 
+  private static final Logger log = LoggerFactory.getLogger(ContinuousVerify.class);
+
   public static class CMapper extends Mapper<Key,Value,LongWritable,VLongWritable> {
 
-    private static final Logger log = LoggerFactory.getLogger(CMapper.class);
+    private static final Logger cMapperLogger = LoggerFactory.getLogger(CMapper.class);
     private final LongWritable row = new LongWritable();
     private final LongWritable ref = new LongWritable();
     private final VLongWritable vrow = new VLongWritable();
@@ -72,9 +74,9 @@ public class ContinuousVerify extends Configured implements Tool {
       } catch (ContinuousWalk.BadChecksumException bce) {
         context.getCounter(Counts.CORRUPT).increment(1L);
         if (corrupt < 1000) {
-          log.error("Bad checksum : " + key);
+          cMapperLogger.error("Bad checksum : " + key);
         } else if (corrupt == 1000) {
-          log.error("Too many bad checksums, not printing anymore!");
+          cMapperLogger.error("Too many bad checksums, not printing anymore!");
         }
         corrupt++;
         return;
@@ -190,7 +192,7 @@ public class ContinuousVerify extends Configured implements Tool {
 
       Path outputPath = new Path(outputDir + "/" + job.getJobName());
       TextOutputFormat.setOutputPath(job, outputPath);
-      System.out.println("Results from this run will be stored in " + outputPath);
+      log.info("Results from this run will be stored in {}", outputPath);
 
       job.waitForCompletion(true);
 
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Random;
 import java.util.Set;
 
@@ -32,10 +33,12 @@ import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.hadoop.mapreduce.AccumuloInputFormat;
 import org.apache.accumulo.testing.TestProps;
 import org.apache.hadoop.conf.Configured;
+import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.VLongWritable;
+import org.apache.hadoop.mapred.FileAlreadyExistsException;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.Mapper;
 import org.apache.hadoop.mapreduce.Reducer;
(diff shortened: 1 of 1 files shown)


Manno15:
Changes look good so far. I think having a case for a user inputting 'n' would be nice as currently, it will throw an IllegalStateException instead of job closing cleanly. Screenshot below

image

DomGarguilo:
> Changes look good so far. I think having a case for a user inputting 'n' would be nice as currently, it will throw an IllegalStateException instead of job closing cleanly. Screenshot below

Yea I wasn't too sure how to best handle that case. Closing cleanly would probably be best.

milleruntime:
Does it resume properly if you chose to keep the directory? If not, you could just have it always delete the directory and not have to bother prompting the user.

Manno15:
Currently, If you choose to keep the directory then it will exit the operation. I do think we could add other options such as overwriting the existing directory or create a new directory to hold the results of the new verify. I do think prompting the user is necessary. I do not want to accidentally overwrite my previous verify results and then have to run the entire map-reduce job over again.

DomGarguilo:
One way to handle this which might be nice would be by default, create a new directory for each verify run. That way the user would not have to be prompted and no results would be lost.

Manno15:
> One way to handle this which might be nice would be by default, create a new directory for each verify run. That way the user would not have to be prompted and no results would be lost.

The only concern there is how large each directory could get when running verify on a larger cluster. I think prompting the user for each run is the safest approach.

DomGarguilo:
> The only concern there is how large each directory could get when running verify on a larger cluster. I think prompting the user for each run is the safest approach.

Good point. Then maybe the user could be prompted to choose from three options:
1. Stop the operation
2. Delete existing directory and rerun
3. Create a new directory in which to store the results

Manno15:
Those options seem reasonable to me.

Manno15:
After further investigation, it does seem safe to just delete the directory after each run or before the next run. Though a prompt to the user isn't a terrible idea, it might not be necessary.

DomGarguilo:
> After further investigation, it does seem safe to just delete the directory after each run or before the next run. Though a prompt to the user isn't a terrible idea, it might not be necessary.

I agree. It looks like the files produced by running verify are empty temporary files that don't have any use. To me it seems like the best course of action would to be to delete the directory before each run without a prompt to the user. I'll work on adding those changes.

keith-turner:
>  It looks like the files produced by running verify are empty temporary files that don't have any use.

When the verify job does not detect any missing data the files are empty.  When missing data is detected, the files contain information about the missing data that is useful in debugging the problem of why Accumulo lost data.

>  To me it seems like the best course of action would to be to delete the directory before each run without a prompt to the user.

I like that behavior.

DomGarguilo:
The changes in 9160790 deletes the output directory (if exists) before each run. With this, the user is not prompted and `verify` can be run multiple times in succession.

Manno15 (approved):
Changes look good. Confirmed that directory gets removed before each run.

DomGarguilo:
It was suggested that a unique directory could be created for each run. This would allow for multiple runs to be compared to one another. I will push some changes that make a new sub-directory who's name reflects the unique name of the mapreduce job carried out by `verify`.

DomGarguilo:
With the changes in 3e6926d, a new subdirectory gets created for each run:
```
dgarguilo@thor:~/github/fluo-uno$ hadoop fs -ls /tmp/ci-verify
Found 5 items
drwxr-xr-x   - dgarguilo supergroup          0 2021-09-14 15:14 /tmp/ci-verify/ContinuousVerify_ci_1631646755241
drwxr-xr-x   - dgarguilo supergroup          0 2021-09-14 15:18 /tmp/ci-verify/ContinuousVerify_ci_1631646992174
drwxr-xr-x   - dgarguilo supergroup          0 2021-09-14 15:33 /tmp/ci-verify/ContinuousVerify_ci_1631647887518
drwxr-xr-x   - dgarguilo supergroup          0 2021-09-14 15:38 /tmp/ci-verify/ContinuousVerify_ci_1631648160103
drwxr-xr-x   - dgarguilo supergroup          0 2021-09-14 15:44 /tmp/ci-verify/ContinuousVerify_ci_1631648563940
```
I also added a print statement letting the user know where the results will be stored.

Manno15:
Your last two commits are interesting. In d828f3d7b57535d341748c12ebbbd73ee5043b1f, you reverted the logging changes but 4b2ee0a brought those back. Was that intended?

DomGarguilo:
> Your last two commits are interesting. In [d828f3d](https://github.com/apache/accumulo-testing/commit/d828f3d7b57535d341748c12ebbbd73ee5043b1f), you reverted the logging changes but [4b2ee0a](https://github.com/apache/accumulo-testing/commit/4b2ee0a402e73f91f9489209738e507a7ec3d428) brought those back. Was that intended?

Yes. I initially removed the logger since it was unused and then realized it would be helpful to have and use it so I added it back.

Manno15 (approved):
New changes look good.

PR #155, opened 2021-09-20
Fix performance test NumberFormatException bug

### LINKED ISSUE #154: ConditionalMutationsPT fails with NumberFormatException
When running the performance tests (via `./bin/performance run`), ConditionalMutationsPT errors out with the following:
```
Exception in thread "main" java.lang.NumberFormatException: For input string: "∞"
	at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
	at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
	at java.base/java.lang.Double.parseDouble(Double.java:543)
	at org.apache.accumulo.testing.performance.tests.ConditionalMutationsPT.runConditionalMutationsTest(ConditionalMutationsPT.java:104)
	at org.apache.accumulo.testing.performance.tests.ConditionalMutationsPT.runTest(ConditionalMutationsPT.java:76)
	at org.apache.accumulo.testing.performance.impl.PerfTestRunner.main(PerfTestRunner.java:51)
```

I think the issue is occuring here:
https://github.com/apache/accumulo-testing/blob/d4151644a9102ccd447a60a402aaeeab748ef230/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java#L153
where a nanosecond value is being converted to a second value. The error happens when the two nanosecond values, `t1` and `t2`, have a difference of less than 1000000000 (1 second). `TimeUnit.NANOSECONDS.toSeconds` will convert the nanos to the nearest second rounding down. So anything less than 1000000000 (1 second) will result in 0 seconds which, in this case, is in the denominator which returns "∞" and throws an error when being parsed.

I think this can be corrected by manually converting from nanos to seconds.

DomGarguilo:
Closes #154 

* Manually convert from nanoseconds to seconds to preserve precision of timing.
* Fixed average for first set of 1-19 runs

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
@@ -276,7 +276,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
       rateSum += randomizeBatchWriteAndScanTime(bw, bs, i);
     }
 
-    reportBuilder.result("avgRate: 21-41",
+    reportBuilder.result("avgRate: 21-40",
         Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
         "RandomizeBatchScanAndWriteTest: average rate (conditions/sec) to write and scan sequence 21-40 post flush");
   }
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
@@ -97,7 +97,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     conditionalMutationsTime(cw, 0);
 
     double rateSum = 0.0;
-    for (long i = 1; i < 21; i++) {
+    for (long i = 1; i <= 20; i++) {
       rateSum += conditionalMutationsTime(cw, i);
     }
 
@@ -108,7 +108,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0.0;
-    for (long i = 21; i < 41; i++) {
+    for (long i = 21; i <= 40; i++) {
       rateSum += conditionalMutationsTime(cw, i);
     }
 
@@ -173,7 +173,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     randomizeConditionalMutationsTime(cw, 0);
 
     double rateSum = 0;
-    for (long i = 1; i < 21; i++) {
+    for (long i = 1; i <= 20; i++) {
       rateSum += randomizeConditionalMutationsTime(cw, i);
     }
 
@@ -184,7 +184,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0;
-    for (long i = 21; i < 41; i++) {
+    for (long i = 21; i <= 40; i++) {
       rateSum += randomizeConditionalMutationsTime(cw, i);
     }
 
@@ -261,7 +261,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     double rateSum = 0;
 
-    for (long i = 1; i < 21; i++) {
+    for (long i = 1; i <= 20; i++) {
       rateSum += randomizeBatchWriteAndScanTime(bw, bs, i);
     }
 
@@ -272,7 +272,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0;
-    for (long i = 21; i < 41; i++) {
+    for (long i = 21; i <= 40; i++) {
       rateSum += randomizeBatchWriteAndScanTime(bw, bs, i);
     }
 
--- a/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
@@ -101,7 +101,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     }
 
     reportBuilder.result("avgRate: 1-19",
-        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
+        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 19)),
         "ConditionalMutationsTest: average rate (conditions/sec) to run sequence 1-19");
 
     env.getClient().tableOperations().flush(tableName, null, null, true);
@@ -150,7 +150,9 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
+    double nanosPerSec = (double) TimeUnit.SECONDS.toNanos(1);
+    // return number of conditions per second
+    return 10000.0 / ((t2 - t1) / nanosPerSec);
   }
 
   private static void runRandomizeConditionalMutationsTest(Environment env, String tableName,
@@ -173,7 +175,7 @@ public class ConditionalMutationsPT implements PerformanceTest {
     }
 
     reportBuilder.result("avgRate: 1-19",
-        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
+        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 19)),
         "RandomizeConditionalMutationsTest: average rate (conditions/sec)  to run sequence 1-19");
 
     env.getClient().tableOperations().flush(tableName, null, null, true);
@@ -233,7 +235,9 @@ public class ConditionalMutationsPT implements PerformanceTest {
 
     long t2 = System.nanoTime();
 
-    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
+    double nanosPerSec = (double) TimeUnit.SECONDS.toNanos(1);
+    // return number of conditions per second
+    return 10000.0 / ((t2 - t1) / nanosPerSec);
   }
 
   private static void runRandomizeBatchScanAndWriteTest(Environment env, String tableName,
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -150,7 +150,9 @@ public static double conditionalMutationsTime(ConditionalWriter cw, long seq) th
 
     long t2 = System.nanoTime();
 
-    return 10000.0 / TimeUnit.NANOSECONDS.toSeconds(t2 - t1);
+    double nanosPerSec = (double) TimeUnit.SECONDS.toNanos(1);
+    // return number of conditions per second
+    return 10000.0 / ((t2 - t1) / nanosPerSec);
```
ctubbsii, line 155: This would be more readable in a new method that can be reused more readably:

```java
return rate(numEvents, startNanos, stopNanos);
```

The method could look something like:

```java
private static double rate(long numEvents, long startNanos, long stopNanos) {
  double nanosPerSec = 1.0E9; // or (double) TimeUnit.SECONDS.toNanos(1)
  double seconds = (t2 - t1) / nanosPerSec;
  return numEvents / seconds;
}
```

ctubbsii (changes requested):

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -101,7 +101,7 @@ private static void runConditionalMutationsTest(Environment env, String tableNam
     }
 
     reportBuilder.result("avgRate: 1-19",
-        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
+        Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 19)),
```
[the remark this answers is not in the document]
keith-turner: > It's weird that we run it 20 times, once before the loop of 19, but not collecting the sum for that first run.

I suspect the first run is done to warm up the JVMs (like make the JIT kick in) and its time is ignored.  Then the times from the next 19 runs are used.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
EdColeman: Alternatively, why not perform the calculation in nanos and convert the rate at the end?

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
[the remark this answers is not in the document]
Manno15: That seems reasonable. Do we do that for other Performance Tests?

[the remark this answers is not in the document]
ctubbsii: Warming up seems reasonable. I just don't know why we have 19 subsequent runs. It might make sense to just round the number of runs off to 20, rather than use 19 as a divisor, which is a little confusing. The warm-up run would benefit from an inline comment like `// warm-up run`

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
ctubbsii: > Alternatively, why not perform the calculation in nanos and convert the rate at the end?

Because then you'd risk losing precision scaling down when you divide by the nanosecond duration only to multiply by 1E9 to scale back up to seconds. And also, it's much more readable to simply compute the number of seconds, and literally return `numEvents / seconds` to get the number of events per second, rather than convert `events/ns` to `events/s` by multiplying by `ns/s`.

For reference, the code that works in `events/ns` first:

```java
private static double rate(long numEvents, long startNanos, long stopNanos) {
  double eventsPerNano = numEvents / ((double) t2 - t1);
  double nanosPerSec = 1.0E9; // or (double) TimeUnit.SECONDS.toNanos(1);
  return eventsPerNano * nanosPerSec;
}
```

Every line that differs here from the previous suggestion is more confusing. My earlier suggestion was substantially more straight-forward.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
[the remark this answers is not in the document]
DomGarguilo: I'm not sure why the first set only runs 19 times, I just thought I would fix the average calculation. I can go through and bump all the numbers up by 1 so that all runs are performed 20 times for consistency.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
DomGarguilo: What about a method to just convert from nanos to seconds? That might make things even more readable.

DomGarguilo: Addressed in dc41ea0 and 27f49d6

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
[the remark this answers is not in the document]
DomGarguilo: Addressed in dc41ea0

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -93,34 +93,37 @@ private static void runConditionalMutationsTest(Environment env, String tableNam
     ConditionalWriter cw = env.getClient().createConditionalWriter(tableName,
         new ConditionalWriterConfig());
 
+    // warm-up run
     conditionalMutationsTime(cw, 0);
 
     double rateSum = 0.0;
-    for (long i = 1; i < 20; i++) {
+    for (long i = 1; i < 21; i++) {
```
ctubbsii, line 100: I think it might be intuitive if `<=` were used:
```suggestion
    for (long i = 1; i <= 20; i++) {
```

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -93,34 +93,37 @@ private static void runConditionalMutationsTest(Environment env, String tableNam
...
+    reportBuilder.result("avgRate: 1-20",
         Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
-        "ConditionalMutationsTest: average rate (conditions/sec) to run sequence 1-19");
+        "ConditionalMutationsTest: average rate (conditions/sec) to run sequence 1-20");
 
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0.0;
-    for (long i = 20; i < 40; i++) {
+    for (long i = 21; i < 41; i++) {
```
ctubbsii, line 111:
```suggestion
    for (long i = 21; i <= 40; i++) {
```

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -165,42 +169,45 @@ private static void runRandomizeConditionalMutationsTest(Environment env, String
     ConditionalWriter cw = env.getClient().createConditionalWriter(tableName,
         new ConditionalWriterConfig());
 
+    // warm-up run
     randomizeConditionalMutationsTime(cw, 0);
 
     double rateSum = 0;
-    for (long i = 1; i < 20; i++) {
+    for (long i = 1; i < 21; i++) {
```
ctubbsii, line 176:
```suggestion
    for (long i = 1; i <= 20; i++) {
```

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -165,42 +169,45 @@ private static void runRandomizeConditionalMutationsTest(Environment env, String
...
+    reportBuilder.result("avgRate: 1-20",
         Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
-        "RandomizeConditionalMutationsTest: average rate (conditions/sec)  to run sequence 1-19");
+        "RandomizeConditionalMutationsTest: average rate (conditions/sec) to run sequence 1-20");
 
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0;
-    for (long i = 20; i < 40; i++) {
+    for (long i = 21; i < 41; i++) {
```
ctubbsii, line 187:
```suggestion
    for (long i = 21; i <= 40; i++) {
```

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -248,43 +256,46 @@ private static void runRandomizeBatchScanAndWriteTest(Environment env, String ta
     BatchWriter bw = env.getClient().createBatchWriter(tableName, new BatchWriterConfig());
     BatchScanner bs = env.getClient().createBatchScanner(tableName, Authorizations.EMPTY, 1);
 
+    // warm-up run
     randomizeBatchWriteAndScanTime(bw, bs, 0);
 
     double rateSum = 0;
 
-    for (long i = 1; i < 20; i++) {
+    for (long i = 1; i < 21; i++) {
```
ctubbsii, line 264:
```suggestion
    for (long i = 1; i <= 20; i++) {
```

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
```diff
@@ -248,43 +256,46 @@ private static void runRandomizeBatchScanAndWriteTest(Environment env, String ta
...
+    reportBuilder.result("avgRate: 1-20",
         Double.parseDouble(new DecimalFormat("#0.00").format(rateSum / 20)),
-        "RandomizeBatchScanAndWriteTest: average rate (conditions/sec)  to write and scan sequence 1-19");
+        "RandomizeBatchScanAndWriteTest: average rate (conditions/sec) to write and scan sequence 1-20");
 
     env.getClient().tableOperations().flush(tableName, null, null, true);
 
     rateSum = 0;
-    for (long i = 20; i < 40; i++) {
+    for (long i = 21; i < 41; i++) {
```
ctubbsii, line 275:
```suggestion
    for (long i = 21; i <= 40; i++) {
```

ctubbsii (approved):
I think `<= 20` is more intuitive than `< 21`, but it's your call whether to include those changes or not. This is ready with or without that improvement.

src/main/java/org/apache/accumulo/testing/performance/tests/ConditionalMutationsPT.java
DomGarguilo: Good idea, definitely more intuitive.

PR #166, opened 2021-10-25
Add deletes to continuous ingest

### LINKED ISSUE #20: Make continuous ingest delete data
apache/accumulo#537 was a WAL recovery bug where deleted data could possibly come back.  This flaw was not found through testing.  It would be nice to delete data in the continuous ingest test to cover this case.  This could be done by periodically deleting a previously written set of linked list.  The linked list would need to be deleted in reverse order to avoid false positives in the test. Could do something like the following.

```
while (true) {
   //write 1,000,000 linked list of 25 nodes
   if(random.nextInt(10) == 0) {
       //delete previously written 1,000,000 linked list of 25 nodes in reverse order
   }
```

DomGarguilo:
Fixes #20 

* Periodically deletes a previously written set of linked list during continuous ingest

### DIFF
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -82,11 +82,9 @@ test.ci.ingest.pause.wait.max=180
 test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
-# Entries will be periodically deleted during ingest if set to true
-test.ci.ingest.delete.enabled=false
 # The probability (in percent) that a set of entries will be deleted during continuous ingest
-# Note: test.ci.ingest.delete must be set to true for this value to be used
-test.ci.ingest.delete.probability=10
+# To disable deletes, set probability to 0
+test.ci.ingest.delete.probability=50
 
 # Batch walker
 # ------------
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -82,9 +82,9 @@ test.ci.ingest.pause.wait.max=180
 test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
-# The probability (in percent) that a set of entries will be deleted during continuous ingest
+# The probability (between 0 and 1) that a set of entries will be deleted during continuous ingest
 # To disable deletes, set probability to 0
-test.ci.ingest.delete.probability=10
+test.ci.ingest.delete.probability=0.1
 
 # Batch walker
 # ------------
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -82,7 +82,7 @@ test.ci.ingest.pause.wait.max=180
 test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
-# The probability (between 0 and 1) that a set of entries will be deleted during continuous ingest
+# The probability (between 0.0 and 1.0) that a set of entries will be deleted during continuous ingest
 # To disable deletes, set probability to 0
 test.ci.ingest.delete.probability=0.1
 
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -82,6 +82,8 @@ test.ci.ingest.pause.wait.max=180
 test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
+# Entries will be periodically deleted during ingest if set to true
+test.ci.ingest.delete=true
 
 # Batch walker
 # ------------
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -83,7 +83,10 @@ test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
 # Entries will be periodically deleted during ingest if set to true
-test.ci.ingest.delete=true
+test.ci.ingest.delete.enabled=false
+# The probability (in percent) that a set of entries will be deleted during continuous ingest
+# Note: test.ci.ingest.delete must be set to true for this value to be used
+test.ci.ingest.delete.probability=10
 
 # Batch walker
 # ------------
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -84,7 +84,7 @@ test.ci.ingest.pause.duration.min=60
 test.ci.ingest.pause.duration.max=120
 # The probability (in percent) that a set of entries will be deleted during continuous ingest
 # To disable deletes, set probability to 0
-test.ci.ingest.delete.probability=50
+test.ci.ingest.delete.probability=10
 
 # Batch walker
 # ------------
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
@@ -66,8 +66,13 @@ public class ContinuousIngest {
   }
 
   private static boolean deletesEnabled(Properties props) {
-    String value = props.getProperty(TestProps.CI_INGEST_DELETE);
-    return Boolean.parseBoolean(value);
+    String stringValue = props.getProperty(TestProps.CI_INGEST_DELETE_ENABLED);
+    return Boolean.parseBoolean(stringValue);
+  }
+
+  private static int getDeleteProbability(Properties props) {
+    String stringValue = props.getProperty(TestProps.CI_INGEST_DELETE_PROBABILITY);
+    return Integer.parseInt(stringValue);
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -175,19 +191,47 @@ public static void main(String[] args) throws Exception {
...
           if (count >= numEntries)
             break out;
+
+          // delete last set of entries in reverse order
+          if (deletePrevious) {
+            log.info("Deleting previous set of entries");
+            while (!mutationStack.empty()) {
+              Mutation m = mutationStack.pop();
+              count--;
+              bw.addMutation(m);
```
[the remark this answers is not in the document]
keith-turner: > We need to flush the batch writer between some of the mutation for safety.

To expand on this.  Currently in this PR the batch writer is being given mutations in reverse order.  However, there must be flush between nodes in the linked that point to each other.  This is because the batch writer may write mutations out in a different order in which it was given and if the process running the batch writer dies then that could look like lost data instead of deleted data.

For example assume we have the nodes NA->N9, N9->N4, and N4->N1.    If the batch writer is asked to delete NA,N9,N4 in that order it may actually delete them in another order like N9,NA,N4.  If the process dies and only N9 is deleted then when running the verification map reduce job we will see that NA->N9 but not see N9 so it looks like data was lost. If we do delete(NA),flush(),delete(N9),flush(),delete(N1) flush then the flushes will make things happen in the order we want.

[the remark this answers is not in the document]
DomGarguilo: I attempted to make these changes in d23b584.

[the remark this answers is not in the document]
DomGarguilo: @keith-turner, I attempted to make these changes in d23b584.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -64,6 +64,14 @@ private static boolean pauseEnabled(Properties props) {
     return Boolean.parseBoolean(value);
   }
 
+  private static int getDeleteProbability(Properties props) {
```
EdColeman, line 67: Did you consider using a floating point value instead of an integer?  Being this is being passed in as a String from props it might not matter - but I think it is more common to be able to specify a probability as a ratio, say 1/10 or 1/2 (or 0.1 or 0.5) - for example [Open Tracing Sampling](https://opentelemetry-python.readthedocs.io/en/latest/sdk/trace.sampling.html)

DomGarguilo: I hadn't considered it but I do like that better than what I current have here. I think I'll change it to a float between 0 and 1 (e.g. 0.4).

DomGarguilo: Addressed in 9e25568

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -191,14 +205,36 @@ public static void main(String[] args) throws Exception {
...
+        // if the previously written entries are scheduled to be deleted
+        if (delete) {
+          log.info("Deleting last portion of written entries");
+          // add delete mutations in the reverse order in which they were written
+          for (int depth = nodeMap.length - 1; depth >= 0; depth--) {
+            for (int index = nodeMap[0].length - 1; index >= 0; index--) {
+              MutationInfo currentNode = nodeMap[depth][index];
+              Mutation m = new Mutation(genRow(currentNode.row));
+              m.putDelete(genCol(currentNode.cf), genCol(currentNode.cq));
+              count--;
```
keith-turner, line 220: Not sure we should decrement this.   Makes it hard to understand what it is happening.  I think a separate delete counter may be better, then we can see number of inserts and deletes that have happened.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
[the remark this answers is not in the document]
keith-turner: @DomGarguilo  the changes look good, I like the new MutationInfo object.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
DomGarguilo: Addressed in 5e0fcf8. I have a few improvements in mind that are unrelated to adding deletes. I plan to make a separate PR after this one gets merged and will add a delete counter there. For now I have removed the decrementing the counter.

conf/accumulo-testing.properties
```diff
@@ -82,6 +82,9 @@ test.ci.ingest.pause.wait.max=180
 test.ci.ingest.pause.duration.min=60
 # Maximum pause duration (in seconds)
 test.ci.ingest.pause.duration.max=120
+# The probability (between 0.0 and 1.0) that a set of entries will be deleted during continuous ingest
+# To disable deletes, set probability to 0
+test.ci.ingest.delete.probability=0.1
```
DomGarguilo, line 87: The last thing I'm not so sure on is what the default value of this should be. Should we have deletes disabled by default?

keith-turner: > The last thing I'm not so sure on is what the default value of this should be. Should we have deletes disabled by default?

People use continuous ingest for different things.  It was created to test the correctness of Accumulo at scale and this is what I think of as its primary use case.  Its secondary use case is that sometimes people run it just to generate load and/or data.  For the primary use case having deletes on by default seems good because it tests more functionality.  For the secondary use case having deletes on may be less desirable.  So since it seems good for primary use case we should probably have something greater than zero. As for what value >0 it should be, I am not sure.  I like the default you have chosen, seems like a good starting point.  We can change it later if we gain some more insight from running with this new feature.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -177,10 +183,12 @@ public static void main(String[] args) throws Exception {
...
           for (int index = 0; index < flushInterval; index++) {
             long rowLong = genLong(rowMin, rowMax, r);
-            byte[] prevRow = genRow(prevRows[index]);
-            prevRows[index] = rowLong;
-            Mutation m = genMutation(rowLong, r.nextInt(maxColF), r.nextInt(maxColQ), cv,
-                ingestInstanceId, count, prevRow, checksum);
+            byte[] prevRow = genRow(nodeMap[depth - 1][index].row);
+            int cfInt = r.nextInt(maxColF);
+            int cqInt = r.nextInt(maxColQ);
+            nodeMap[depth][index] = new MutationInfo(rowLong, cfInt, cqInt);
```
keith-turner, line 189: We may be able to avoid creating these MutationInfo objects when we are not deleting.  Would need to move the creation of the delete var to before this loop.  Less object creation may speed things up a bit when we are not deleting.

```suggestion
            if(delete || depth == maxDepth - 1) {
               nodeMap[depth][index] = new MutationInfo(rowLong, cfInt, cqInt);
            }
```

keith-turner (approved):
@DomGarguilo the changes look really good. Have you done any manual testing of this?   Would be nice to verify the code is doing whats expected, but I am not sure of the best way to do that.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
DomGarguilo: Don't we need these values in order to create the `byte[] prevRow` regardless of whether we are deleting or not?

keith-turner: Yeah we do need them for that, I was only looking at the later code.

DomGarguilo:
> Have you done any manual testing of this?

I have done some testing. Configuring CI to write a full set of 1,000,000 nodes at depth 25 allows the deletes to occur. It seems that all nodes written are then deleted once a compaction happens (I manually compacted).

Something I found while doing that testing is, it is not possible to have all written entries deleted. For example, if I set the entries property to **25,000,000**, then the code will exit the loop ([here](https://github.com/apache/accumulo-testing/blob/ae207b1bd7a855a8abec2fe42c2559d2bf26405b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java#L197-L198)) before reaching the portion that initiates the deletes. If I set the entries to **25,000,001**, then the deletes **will** happen, but another 1M entries will be written before then next check ([here](https://github.com/apache/accumulo-testing/blob/ae207b1bd7a855a8abec2fe42c2559d2bf26405b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java#L179-L180)) will trigger an exit, leaving the total entries at 1M. I don't think this is a big deal, just thought it was interesting and also don't see a clean way to allow this to happen.

Manno15 (approved):
Changes look good!

PR #168, opened 2021-11-19
Continuous Ingest improvements

DomGarguilo:
While working on #166 I noticed quite a few improvements that could be made to ContinuousIngest that were unrelated.

* renamed variables
* made some variables final
* consolidated for loops
* improved logs
* avoided parsing properties multiple times
* move AccumuloClient and BatchWriter into try-with-resources
* created member variables to simplify methods

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
@@ -113,8 +113,6 @@ public class ContinuousIngest {
             "Consult the README and create the table before starting ingest.");
       }
 
-      BatchWriter bw = client.createBatchWriter(tableName);
-
       Random rand = new Random();
 
       byte[] ingestInstanceId = UUID.randomUUID().toString().getBytes(UTF_8);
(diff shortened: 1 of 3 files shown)


DomGarguilo:
One of the more evident improvements that I was hoping to make was to the flush log messages. Currently they look like this: 
```
INFO : FLUSH 1637343614117 6475 782 25000000 1000000
```
Where none of the values have labels so it is hard to tell what is going on. The current changes now produce:
```
INFO : FLUSH - current time: 1637353766960ms, time since last flush: 6569ms, flush duration: 690ms, # of entries written: 36999999, # of entries deleted: 0, flush interval: 1000000
```
which I think is a lot more informative but also a lot longer. I'm not sure if this is the best solution so if anyone has an opinion on how this might be further improved/left alone let me know.

Manno15:
> INFO : FLUSH - current time: 1637353766960ms, time since last flush: 6569ms, flush duration: 690ms, # of entries written: 36999999, # of entries deleted: 0, flush interval: 1000000

I do like the idea of a more informative message. I think the labels could be shortened a bit but will still be descriptive. Something Like: 
```
INFO : FLUSH - current time: 1637353766960ms, last flush: 6569ms, duration: 690ms,  written: 36999999, eleted: 0,  interval: 1000000
```
Also, the large ms value for current time isn't super useful as is (to me anyways). It could be converted to something more human-readable.

DomGarguilo:
Here is an updated look at the logs produced with these changes:
```
2021-11-22T11:41:47,058 [testing.continuous.ContinuousIngest] INFO : Ingest instance ID: 51b4823c-50ec-42c5-bcc9-8dd7b2a1613f current time: 1637599307056ms
2021-11-22T11:41:47,059 [testing.continuous.ContinuousIngest] INFO : A flush will occur after every 1000000 entries written
2021-11-22T11:41:47,094 [testing.continuous.ContinuousIngest] INFO : Total entries to be written: 9223372036854775807
2021-11-22T11:41:47,095 [testing.continuous.ContinuousIngest] INFO : DELETES will occur with a probability of 0.10
2021-11-22T11:41:50,230 [testing.continuous.ContinuousIngest] INFO : FLUSH - duration: 386ms, since last flush: 3136ms, total written: 1000000, total deleted: 0
2021-11-22T11:41:53,747 [testing.continuous.ContinuousIngest] INFO : FLUSH - duration: 315ms, since last flush: 3517ms, total written: 2000000, total deleted: 0
2021-11-22T11:41:57,652 [testing.continuous.ContinuousIngest] INFO : FLUSH - duration: 291ms, since last flush: 3905ms, total written: 3000000, total deleted: 0
2021-11-22T11:42:01,906 [testing.continuous.ContinuousIngest] INFO : FLUSH - duration: 311ms, since last flush: 4254ms, total written: 4000000, total deleted: 0
2021-11-22T11:42:06,588 [testing.continuous.ContinuousIngest] INFO : FLUSH - duration: 541ms, since last flush: 4682ms, total written: 5000000, total deleted: 0
```
This differs from what was logged before by removing the timestamp and moving the flush interval to a single log outside of the loop.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -86,51 +82,49 @@ private static int getFlushEntries(Properties props) {
...
             "Consult the README and create the table before starting ingest.");
       }
 
-      BatchWriter bw = client.createBatchWriter(tableName);
-
-      Random r = new Random();
+      Random rand = new Random();
 
       byte[] ingestInstanceId = UUID.randomUUID().toString().getBytes(UTF_8);
+      log.info("Ingest instance ID: {} current time: {}ms", new String(ingestInstanceId, UTF_8),
```
Manno15, line 119: I like the new log messages but I am not quite sure the usefulness of this one. Especially about current time. It is possible we could shift some of the log message to debug level but considering these only get displayed once per command run, it isn't necessary.

DomGarguilo: Yea I agree, I'm not sure that I see the usefulness of this log, I just left it in since it was already there and made it more descriptive.

keith-turner: Not sure how useful the time is, but the UUID is very useful to know.  When data is lost there will be UUID in the accumulo table pointing to the missing data.  Can use that to fine the ingest logs if those logs contain the UUID.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -141,99 +135,96 @@ public static void main(String[] args) throws Exception {
...
 
-      Properties testProps = env.getTestProperties();
-      if (pauseEnabled(testProps)) {
+      visibilities = parseVisibilities(testProps.getProperty(TestProps.CI_INGEST_VISIBILITIES));
+
+      pauseEnabled = pauseEnabled(testProps);
+
+      pauseMin = Integer.parseInt(testProps.getProperty(TestProps.CI_INGEST_PAUSE_WAIT_MIN));
+      pauseMax = Integer.parseInt(testProps.getProperty(TestProps.CI_INGEST_PAUSE_WAIT_MAX));
+      Preconditions.checkState(pauseMax >= pauseMin && pauseMin > 0);
```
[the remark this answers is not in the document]
DomGarguilo: Good idea, will do.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
DomGarguilo: Gotcha, thanks. I think I'll leave this log as it appears above just in case the timestamp could be useful.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -141,99 +134,97 @@ public static void main(String[] args) throws Exception {
...
-        for (int index = 0; index < flushInterval; index++) {
-          long rowLong = genLong(rowMin, rowMax, r);
+          // generate sets nodes that link to previous set of nodes
+          for (int depth = 0; depth < maxDepth; depth++) {
+            for (int index = 0; index < flushInterval; index++) {
+              long rowLong = genLong(rowMin, rowMax, rand);
 
-          int cf = r.nextInt(maxColF);
-          int cq = r.nextInt(maxColQ);
+              byte[] prevRow = depth == 0 ? null : genRow(nodeMap[depth - 1][index].row);
```
keith-turner, line 174: It seems like adding this check for `depth == 0` allowed you to squish two loops into one.  Is that correct?  Just want to make sure I am understanding this correctly because I was finding it hard to follow the change in the diff for this.

DomGarguilo: Yes that is correct, sorry for the large diff.

keith-turner: These changes are great, sometimes diffs are just hard to grok.  Thanks for the confirmation, you saved me time from looking at the two bits of code outside of a diff.

Manno15 (approved):
Changes look good.

PR #172, opened 2021-11-24
Fix build warnings in a few classes due to deprication

Manno15:
There were a few build warnings due to Class.newInstance() being deprecated. I am not 100% sure if this is the most ideal fix but it does remove those warnings.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/performance/impl/PerfTestRunner.java
+++ b/src/main/java/org/apache/accumulo/testing/performance/impl/PerfTestRunner.java
@@ -41,7 +41,7 @@ public class PerfTestRunner {
     String accumuloVersion = args[2];
     String outputDir = args[3];
 
-    PerformanceTest perfTest = Class.forName(className).asSubclass(PerformanceTest.class)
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/accumulo/testing/performance/impl/PerfTestRunner.java
```diff
@@ -41,7 +41,7 @@ public static void main(String[] args) throws Exception {
     String accumuloVersion = args[2];
     String outputDir = args[3];
 
-    PerformanceTest perfTest = Class.forName(className).asSubclass(PerformanceTest.class)
+    PerformanceTest perfTest = Class.forName(className).asSubclass(PerformanceTest.class).getDeclaredConstructor()
```
milleruntime, line 44: Format
```suggestion
    PerformanceTest perfTest = Class.forName(className).asSubclass(PerformanceTest.class)
        .getDeclaredConstructor().newInstance();
```

milleruntime (approved):

DomGarguilo (approved):

DomGarguilo:
> I am not 100% sure if this is the most ideal fix but it does remove those warnings.

I think this is fine. This is what we did in #151 for the same issue but just for occurrence.

PR #173, opened 2021-11-24
Fix  module-info warnings

Manno15:
A couple more warnings in the build. 

```
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
[WARNING] Discovered module-info.class. Shading will break its strong encapsulation.
```
Currently ran into a bug with this section
<del>
WARNING] zookeeper-3.4.14.jar, zookeeper-jute-3.5.9.jar define 101 overlapping classes: 
[WARNING]   - org.apache.zookeeper.proto.GetMaxChildrenRequest
[WARNING]   - org.apache.jute.compiler.JavaGenerator
[WARNING]   - org.apache.jute.XmlInputArchive$XMLParser
[WARNING]   - org.apache.jute.CsvInputArchive$1
[WARNING]   - org.apache.zookeeper.proto.SetACLRequest
[WARNING]   - org.apache.jute.compiler.JLong
[WARNING]   - org.apache.zookeeper.server.quorum.QuorumPacket
[WARNING]   - org.apache.zookeeper.proto.SetSASLRequest
[WARNING]   - org.apache.zookeeper.data.Id
[WARNING]   - org.apache.zookeeper.proto.SyncRequest
[WARNING]   - 91 more...
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See http://maven.apache.org/plugins/maven-shade-plugin/

</del>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -399,7 +399,6 @@
                   <artifactSet>
                     <excludes>
                       <exclude>org.apache.accumulo:accumulo-native</exclude>
-                      <exclude>org.apache.zookeeper*</exclude>
                       <exclude>org.apache.hadoop:*</exclude>
                       <exclude>log4j:*</exclude>
                     </excludes>
--- a/pom.xml
+++ b/pom.xml
@@ -399,6 +399,7 @@
                   <artifactSet>
                     <excludes>
                       <exclude>org.apache.accumulo:accumulo-native</exclude>
+                      <exclude>org.apache.zookeeper*</exclude>
                       <exclude>org.apache.hadoop:*</exclude>
                       <exclude>log4j:*</exclude>
                     </excludes>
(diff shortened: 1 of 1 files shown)


milleruntime:
I am still seeing the warnings for `[WARNING] zookeeper-3.4.14.jar, zookeeper-jute-3.5.9.jar define 101 overlapping classes: `. But it looks like you changed this to only fix the module-info warnings?

Manno15:
That is correct. I ran into a bug while testing performance tests. So I need to reevaluate it.

milleruntime (approved):

PR #174, opened 2021-11-24
Update zookeeper version to fix Overlapping warning

Manno15:
Build script was still exporting zookeeper version 3.14.4. Updating that to version 3.7.0 fixes the overlap warning while building.

### DIFF
--- a/conf/env.sh
+++ b/conf/env.sh
@@ -61,7 +61,7 @@ fi
 # Versions set below will be what is included in the shaded jar
 ACCUMULO_VERSION="$("$ACCUMULO_HOME"/bin/accumulo version | grep -v 'DEBUG')"; export ACCUMULO_VERSION
 HADOOP_VERSION="$(hadoop version | head -n1 | awk '{print $2}')"; export HADOOP_VERSION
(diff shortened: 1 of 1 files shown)


conf/env.sh
```diff
@@ -61,7 +61,7 @@ fi
 # Versions set below will be what is included in the shaded jar
 ACCUMULO_VERSION="$("$ACCUMULO_HOME"/bin/accumulo version | grep -v 'DEBUG')"; export ACCUMULO_VERSION
 HADOOP_VERSION="$(hadoop version | head -n1 | awk '{print $2}')"; export HADOOP_VERSION
-export ZOOKEEPER_VERSION=3.4.14
+export ZOOKEEPER_VERSION=3.7.0
```
ctubbsii, line 64: Should this use 3.5.9 instead, which is the base version that Accumulo 2.1 was developed with?

Manno15: I didn't realize that. 3.5.9 also fixes the issue, I will update PR.

DomGarguilo (approved):

PR #176, opened 2021-12-07
Some minor code cleanup throughout accumulo-testing

Manno15:


src/main/java/org/apache/accumulo/testing/randomwalk/concurrent/CreateTable.java
```diff
@@ -45,7 +45,7 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
       else
         throw e;
     } catch (IllegalArgumentException e) {
-      log.debug("Create: " + e.toString());
+      log.debug("Create: " + e);
```
DomGarguilo, line 48:
```suggestion
      log.debug("Create: {}", e);
```

src/main/java/org/apache/accumulo/testing/randomwalk/shard/Grep.java
```diff
@@ -88,13 +88,11 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
     bs.close();
 
     if (!documentsFoundInIndex.equals(documentsFoundByGrep)) {
-      throw new Exception(
-          "Set of documents found not equal for words " + Arrays.asList(words).toString() + " "
-              + documentsFoundInIndex + " " + documentsFoundByGrep);
+      throw new Exception("Set of documents found not equal for words " + Arrays.asList(words) + " "
```
DomGarguilo, line 91:
```suggestion
      throw new Exception("Set of documents found not equal for words " + Arrays.toString(words) + " "
```
Does the same thing without having to first convert it to List.

src/main/java/org/apache/accumulo/testing/randomwalk/shard/Grep.java
```diff
@@ -88,13 +88,11 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
...
+          + documentsFoundInIndex + " " + documentsFoundByGrep);
     }
 
-    log.debug("Grep and index agree " + Arrays.asList(words).toString() + " "
-        + documentsFoundInIndex.size());
+    log.debug("Grep and index agree " + Arrays.asList(words) + " " + documentsFoundInIndex.size());
```
DomGarguilo, line 95:
```suggestion
    log.debug("Grep and index agree " + Arrays.toString(words) + " " + documentsFoundInIndex.size());
```
Does the same thing without having to first convert it to List.

milleruntime (approved):

DomGarguilo (approved):

PR #178, opened 2021-12-14
fix javadoc warnings

EdColeman:
Fixes errors when building java docs (mvn clean verify javadoc:jar)

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/TestEnv.java
+++ b/src/main/java/org/apache/accumulo/testing/TestEnv.java
@@ -84,8 +84,6 @@ public class TestEnv implements AutoCloseable {
   }
 
   /**
-   * @param key
-   *          the property name
    * @return a test property value given a key
    */
   public String getTestProperty(String key) {
@@ -155,8 +153,6 @@ public class TestEnv implements AutoCloseable {
 
   /**
    * Gets an authentication token based on the configured password.
-   * 
-   * @return the AuthenticationToken
    */
   public AuthenticationToken getToken() {
     return ClientProperty.getAuthenticationToken(clientProps);
@@ -172,8 +168,6 @@ public class TestEnv implements AutoCloseable {
 
   /**
    * Gets an Accumulo client. The same client is reused after the first call.
-   * 
-   * @return the Accumulo client
    */
   public synchronized AccumuloClient getAccumuloClient() {
     if (client == null) {
--- a/src/main/java/org/apache/accumulo/testing/TestEnv.java
+++ b/src/main/java/org/apache/accumulo/testing/TestEnv.java
@@ -84,6 +84,8 @@ public class TestEnv implements AutoCloseable {
   }
 
   /**
+   * @param key
+   *          the property name
    * @return a test property value given a key
    */
   public String getTestProperty(String key) {
@@ -153,6 +155,8 @@ public class TestEnv implements AutoCloseable {
 
   /**
    * Gets an authentication token based on the configured password.
+   * 
+   * @return the AuthenticationToken
    */
   public AuthenticationToken getToken() {
     return ClientProperty.getAuthenticationToken(clientProps);
@@ -168,6 +172,8 @@ public class TestEnv implements AutoCloseable {
 
   /**
    * Gets an Accumulo client. The same client is reused after the first call.
+   * 
+   * @return the Accumulo client
    */
   public synchronized AccumuloClient getAccumuloClient() {
     if (client == null) {
--- a/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
@@ -251,7 +251,7 @@ public class UndefinedAnalyzer {
    * @param args
    *          the command line arguments
    * @throws Exception
-   *           is an exception occurs.
+   *           if an exception occurs
    */
   public static void main(String[] args) throws Exception {
     Opts opts = new Opts();
--- a/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
@@ -247,11 +247,6 @@ public class UndefinedAnalyzer {
   /**
    * Class to analyze undefined references and accumulo logs to isolate the time/tablet where data
    * was lost.
-   * 
-   * @param args
-   *          the command line arguments
-   * @throws Exception
-   *           if an exception occurs
    */
   public static void main(String[] args) throws Exception {
     Opts opts = new Opts();
--- a/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
@@ -247,6 +247,11 @@ public class UndefinedAnalyzer {
   /**
    * Class to analyze undefined references and accumulo logs to isolate the time/tablet where data
    * was lost.
+   * 
+   * @param args
+   *          the command line arguments
+   * @throws Exception
+   *           is an exception occurs.
    */
   public static void main(String[] args) throws Exception {
     Opts opts = new Opts();
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/Framework.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/Framework.java
@@ -43,7 +43,7 @@ public class Framework {
    * @param env
    *          the test environment
    * @param state
-   *          tRandom walk state passed between nodes
+   *          Random walk state passed between nodes
    *
    * @return 0 if successful, -1 if an exception is thrown during the random walk
    */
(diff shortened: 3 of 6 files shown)


src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
```diff
@@ -247,6 +247,11 @@ TabletAssignment findMostRecentAssignment(String row, long time1, long time2) {
   /**
    * Class to analyze undefined references and accumulo logs to isolate the time/tablet where data
    * was lost.
+   * 
+   * @param args
+   *          the command line arguments
+   * @throws Exception
+   *           is an exception occurs.
```
DomGarguilo, line 254:
```suggestion
   *           if an exception occurs
```

src/main/java/org/apache/accumulo/testing/randomwalk/Framework.java
```diff
@@ -40,6 +40,12 @@ public static Framework getInstance() {
    *
    * @param startName
    *          Full name of starting graph or test
+   * @param env
+   *          the test environment
+   * @param state
+   *          tRandom walk state passed between nodes
```
DomGarguilo, line 46:
```suggestion
   *          Random walk state passed between nodes
```

src/main/java/org/apache/accumulo/testing/randomwalk/RandWalkEnv.java
```diff
@@ -47,6 +47,10 @@ public RandWalkEnv(String testPropsPath, String clientPropsPath) {
    *           if any of the numeric batch writer configuration properties cannot be parsed
    * @throws NumberFormatException
    *           if any configuration property cannot be parsed
+   * @throws AccumuloException
+   *           if an Accumulo exception occurs
+   * @throws AccumuloSecurityException
+   *           is an Accumulo security exception occurs.
```
DomGarguilo, line 53:
```suggestion
   *           if an Accumulo security exception occurs
```

src/main/java/org/apache/accumulo/testing/randomwalk/State.java
```diff
@@ -107,6 +112,8 @@ public String getRandomNamespace() {
 
   /**
    * Gets a random table name
+   * 
+   * @return a random table name from current table names.
```
DomGarguilo, line 116:
```suggestion
   * @return a random table name from the list of current table names.
```

DomGarguilo (approved):
I was unable to reproduce the warnings for some reason but these changes LGTM regardless.

src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
```diff
@@ -247,6 +247,11 @@ TabletAssignment findMostRecentAssignment(String row, long time1, long time2) {
   /**
    * Class to analyze undefined references and accumulo logs to isolate the time/tablet where data
    * was lost.
+   * 
+   * @param args
+   *          the command line arguments
+   * @throws Exception
+   *           if an exception occurs
```
ctubbsii, line 254: I think you have your warnings set weird. We normally exclude missing standard javadocs from warnings, because it forces us to put things in like this, which doesn't add any value from what you can see already from the method signature.

ctubbsii (commented):
Some of these add value, but a lot of them just repeat information already shown in the method signatures, which isn't useful. I updated the javadoc configuration in #180 so the Maven build doesn't produce javadoc warnings on missing standard tags, so we can add them when they are useful, but don't need to if they aren't. Of course, your IDE may still show its own warnings if it doesn't respect the Maven plugin configuration. I use Eclipse, which doesn't warn on javadocs by default. I turn on the javadoc warnings, but turn off the warnings for missing standard tags.

src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
EdColeman: For the record - I saw the errors in the maven build - it was not a setting in my IDE.

If the update to the configuration eliminates the warning, this PR can be rejected.

EdColeman: Verified that the changes in #180 eliminates the warnings - the changes in this PR are not necessary to remove the warnings when building with maven.

ctubbsii: Cool, but if you want to keep any of the updated javadocs, feel free to do so. Some of these could still be useful for their explanatory power, and not just to eliminate warnings.

src/main/java/org/apache/accumulo/testing/TestEnv.java
```diff
@@ -84,6 +84,8 @@ public Properties getTestProperties() {
   }
 
   /**
+   * @param key
+   *          the property name
```
ctubbsii, line 88:
```suggestion
```

src/main/java/org/apache/accumulo/testing/TestEnv.java
```diff
@@ -153,6 +155,8 @@ public Configuration getHadoopConfiguration() {
 
   /**
    * Gets an authentication token based on the configured password.
+   * 
+   * @return the AuthenticationToken
```
ctubbsii, line 159:
```suggestion
```

src/main/java/org/apache/accumulo/testing/TestEnv.java
```diff
@@ -168,6 +172,8 @@ public String getYarnResourceManager() {
 
   /**
    * Gets an Accumulo client. The same client is reused after the first call.
+   * 
+   * @return the Accumulo client
```
ctubbsii, line 176:
```suggestion
```

src/main/java/org/apache/accumulo/testing/TestEnv.java
ctubbsii, line 88:
```suggestion
```

src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
ctubbsii:
```suggestion
```

src/main/java/org/apache/accumulo/testing/randomwalk/Node.java
```diff
@@ -36,6 +36,10 @@
    *          Random walk state passed between nodes
    * @param env
    *          test environment
+   * @param props
+   *          properties
+   * @throws Exception
+   *           if an exception occurs
```
ctubbsii, line 42:
```suggestion
```

src/main/java/org/apache/accumulo/testing/randomwalk/RandWalkEnv.java
```diff
@@ -47,6 +47,10 @@ public RandWalkEnv(String testPropsPath, String clientPropsPath) {
    *           if any of the numeric batch writer configuration properties cannot be parsed
    * @throws NumberFormatException
    *           if any configuration property cannot be parsed
+   * @throws AccumuloException
+   *           if an Accumulo exception occurs
+   * @throws AccumuloSecurityException
+   *           if an Accumulo security exception occurs
```
ctubbsii, line 53:
```suggestion
```

src/main/java/org/apache/accumulo/testing/randomwalk/State.java
```diff
@@ -51,13 +51,18 @@ public void set(String key, Object value) {
 
   /**
    * Resets the Random object.
+   * 
+   * @param rand
+   *          a random number generator.
```
ctubbsii, line 56:
```suggestion
```

src/main/java/org/apache/accumulo/testing/randomwalk/State.java
```diff
@@ -51,13 +51,18 @@ public void set(String key, Object value) {
...
    */
   public void setRandom(Random rand) {
     this.random = rand;
   }
 
   /**
    * Gets the random object.
+   * 
+   * @return the random generator
```
ctubbsii, line 65:
```suggestion
```

ctubbsii (commented):
Now that the warnings are gone in #180, the following suggested changes roll back the added javadoc tags that were there to remove warnings. I attempted to keep the ones that seemed like useful explanatory additions.

src/main/java/org/apache/accumulo/testing/continuous/UndefinedAnalyzer.java
EdColeman: The comment 
```Some of these add value, but a lot of them just repeat information already shown in the method signatures, which isn't useful. ```
is subjective - the changes I added were from a simple reading of the signature / or the code, not from a deep understanding - so I could argue that they add all add very little information.

File: src/main/java/org/apache/accumulo/testing/randomwalk/State.java [hunk 13]
```diff
@@ -107,6 +112,8 @@ public String getRandomNamespace() {
 
   /**
    * Gets a random table name
+   *
```
ctubbsii, line 115: Removing whitespace

```suggestion
   *
```

ctubbsii (approved):

PR #179, opened 2021-12-14
Replace use of `pssh` with `parallel-ssh`

DomGarguilo:
* Replace use of `pssh` with `parallel-ssh`
* Replace use of `pscp` with `parallel-scp`

I ran into the following issue when trying to run `./bin/agitator start`:
```
The agitator requires pssh to be installed. Aborting.
```
Which comes from here: https://github.com/apache/accumulo-testing/blob/fbaf16dc2c065eaab690d8d7abd75750d5d0425c/bin/agitator#L34

After looking into it I found this in `/usr/share/doc/pssh/README.Debian`:
```
To avoid any conflicts with the putty package, all of the programs have been
renamed.

parallel-ssh is pssh
parallel-scp is pscp
parallel-rsync is prsync
parallel-nuke is pnuke
parallel-slurp is pslurp
```

After replacing, I haven't run into these issues.


EdColeman:
Is the rename universal for different linux distributions?  With a quick look, [redhat / centos](https://www.cyberithub.com/install-parallel-ssh-on-redhat-centos-7/) may still be using pssh.  An alternative could be to modify the script(s) to it tries parallel-ssh and if not found tries pssh. But I am unsure how universal we need to make these scripts.

DomGarguilo:
> Is the rename universal for different linux distributions?

Not too sure. I can look into that.

DomGarguilo:
> Fedora/CentOS/RHEL has not renamed this binary. It still ships as pssh.rpm and installs a binary at /usr/bin/pssh.
> 
> An alternative to this would be to set a variable, `$PSSH` equal to `pssh` by default, and then you can set `export PSSH=parallel-ssh` in your own environment to override.

Gotcha, maybe this PR could add a note to the log stmts and/or Readme instead if that seems like it would be helpful.

ctubbsii:
> > Fedora/CentOS/RHEL has not renamed this binary. It still ships as pssh.rpm and installs a binary at /usr/bin/pssh.
> > An alternative to this would be to set a variable, `$PSSH` equal to `pssh` by default, and then you can set `export PSSH=parallel-ssh` in your own environment to override.
> 
> Gotcha, maybe this PR could add a note to the log stmts and/or Readme instead, explaining why this might be failing, if that seems like it would be helpful.

If you change the implementation to provide the binary in a variable, like `$PSSH`, it would be totally appropriate to update the log message to indicate that `pssh` could not be found and that `$PSSH` is not set to an equivalent executable that exists, and to update the README to explain the requirement and how to set the `$PSSH` variable.

Another alternative might be to automatically fall back to `parallel-ssh` if that [binary exists on the `$PATH` when `pssh` doesn't](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script).

DomGarguilo:
> Another alternative might be to automatically fall back to `parallel-ssh` if that [binary exists on the `$PATH` when `pssh` doesn't](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script).

That seems reasonable. I put the following code in env.sh 
```bash
if hash pssh 2>/dev/null; then
  export PSSH=pssh
elif hash parallel-ssh 2>/dev/null; then
  export PSSH=parallel-ssh
else
  echo >&2 "Some components require pssh/parallel-ssh to be installed. Aborting."; exit 1;
fi
```
and then replaced usages of `pssh` with `$PSSH` which seems to work well. If that seems like a reasonable approach I can push those changes. It seems like we either put the logic for this in one common spot (i.e. env.sh) or in every script that needs pssh.

DomGarguilo:
In 0807c98 I added checks in two places,  [the agitator script](https://github.com/apache/accumulo-testing/blob/0807c98b67c6fa18e0fd73a07f5216f3663a9479/bin/agitator) and [stress-env.sh](https://github.com/apache/accumulo-testing/blob/0807c98b67c6fa18e0fd73a07f5216f3663a9479/test/stress/stress-env.sh). These two places will assign either `pssh` or `parallel-ssh` to `$PSSH` based on which is available. The rest of the occurances of `pssh` have been replaced with `$PSSH`.

ctubbsii (approved):

libexec/master-agitator.pl
```diff
@@ -30,6 +30,12 @@
   print "ERROR: ACCUMULO_HOME needs to be set!";
   exit(1);
 }
+if( defined $ENV{'PSSH'} ){
+  $PSSH = $ENV{'PSSH'};
+} else {
+  print "ERROR: PSSH needs to be set!";
+  exit(1);
+}
```
milleruntime, line 38: I am not sure why but this error is being printed for me. Both commands `pssh` and `parallel-ssh` work on my machine.

milleruntime: Running `export PSSH=parallel-ssh` will solve this but I thought the changes in `start_agitator()` in the agitator script were supposed to automatically set the env variable.

ctubbsii: Oops, the variable needs to be exported to be seen by the perl script. Should be fixed in 6a665e7a1c995c0f7f40aa1dfde29d515ed706fe

PR #184, opened 2022-01-06
Fix broken agitator scripts

### LINKED ISSUE #183: Update agitator to read cluster.yml
This ticket is for a solution to have the agitator scripts read the new cluster.yml file. The files referenced in the agitator scripts (master-agitator.pl, tserver-agitator.pl, datanode-agitator.pl) no longer exist since https://github.com/apache/accumulo/pull/2238. Related to #182 

One solution would be to use the perl library "tiny" to real the YAML file but this will require the install of an additional perl package. For example, if the test cluster were running centos it would need to have the package [perl-yaml-tiny](https://centos.pkgs.org/7/centos-x86_64/perl-YAML-Tiny-1.51-6.el7.noarch.rpm.html) installed. I am not sure if this is the best solution.

milleruntime:
* Create bash functions to start agitators
* Drop broken perl scripts in favor of bash functions
* Add function to call ClusterConfigParser to read cluster.yaml
* Make agitators only log to one file
* Closes #183


milleruntime:
FYI I am currently working through `shellcheck` recommendations so please hold off on any of that type of feedback.

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
 
+function run_agitators() {
+  unset APP_HOSTS
+  export APP_HOSTS=$MANAGER_HOSTS
+  echo "Starting manager agitation as ${1} for ${APP_HOSTS}"
+  start_app_agitator "manager" "${AGTR_MASTER_KILL_SLEEP_TIME}" "${AGTR_MASTER_RESTART_SLEEP_TIME}" 1 1 > "${manager_log}" 2>&1 &
+  unset APP_HOSTS
+  export APP_HOSTS=$TSERVER_HOSTS
+  echo "Starting tserver agitation as ${1} for ${APP_HOSTS}"
+  start_app_agitator "tserver" "${AGTR_TSERVER_KILL_SLEEP_TIME}" "${AGTR_TSERVER_RESTART_SLEEP_TIME}" "${AGTR_TSERVER_MIN_KILL}" "${AGTR_TSERVER_MAX_KILL}" > "${tserver_log}" 2>&1 &
```
ctubbsii, line 41: These first three lines are duplicated with different parameters for the managers and the tservers. These can be passed as parameters to the `start_app_agitator` function instead.

Also, if at all possible, you should avoid passing around variables in and out of functions using global variables. This is especially important since you're backgrounding the calls and then immediately editing the value of the global variable, which could create concurrency issues. You can declare variables local to a function with the `local` keyword, as in `local x; x=$1` to assign the first argument to the function to the local variable named `x`.

Also, you don't need all the curly braces for scalar variables that aren't ambiguous in a string. They can add noise that affect readability.

```suggestion
  local agtr_user=$1
  start_app_agitator manager "$agtr_user" "$AGTR_MASTER_KILL_SLEEP_TIME" "$AGTR_MASTER_RESTART_SLEEP_TIME" 1 1 $MANAGER_HOSTS > "$manager_log" 2>&1 &
  start_app_agitator tserver "$agtr_user" "$AGTR_TSERVER_KILL_SLEEP_TIME" "$AGTR_TSERVER_RESTART_SLEEP_TIME" "$AGTR_TSERVER_MIN_KILL" "$AGTR_TSERVER_MAX_KILL" $TSERVER_HOSTS > "$tserver_log" 2>&1 &
```

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
+
+# Starts a app specific agitator
+# usage: start_app_agitator app_name kill_sleep_time restart_sleep_time min_kill max_kill
+# Requires that a list of hosts to be set in $APP_HOSTS
+function start_app_agitator() {
+  app_name=$1
+  kill_sleep_time=$2
+  restart_sleep_time=$3
+  min_kill=$4
+  max_kill=$5
```
ctubbsii, line 52:
```suggestion
# Requires that a list of hosts to be set in $APP_HOSTS
function start_app_agitator() {
  local app_name=$1; shift
  local agtr_user=$1; shift
  local kill_sleep_time=$1; shift
  local restart_sleep_time=$1; shift
  local min_kill=$1; shift
  local max_kill=$1; shift
  local APP_HOSTS=("$@") # can keep this as an array
  echo "Starting $app_name agitation as $agtr_user for ${APP_HOSTS[*]}"
```

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
+  ENV_VARS="ACCUMULO_HOME=${ACCUMULO_HOME} ZOOKEEPER_HOME=${ZOOKEEPER_HOME} HADOOP_HOME=${HADOOP_HOME} JAVA_HOME=${JAVA_HOME}"
+  if [[ -z $APP_HOSTS ]]; then
+    echo "ERROR: No hosts were found in env for ${app_name}"
+    exit 1
+  fi
```
ctubbsii, line 57: Based on my other suggestions, this could just check the size of the array

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
+  T="$(date +'%Y%m%d %H:%M:%S')"
+  echo "${T} Running ${app_name} agitator ${kill_sleep_time} ${restart_sleep_time} ${min_kill} ${max_kill} for HOSTS=${APP_HOSTS}"
+  while true; do
+    echo "${T} Sleeping for ${kill_sleep_time} minutes"
+    sleep $((kill_sleep_time * 60))
+
+    T="$(date +'%Y%m%d %H:%M:%S')"
+    numToWack=$((min_kill + RANDOM % max_kill))
+    nodeToWack=$(echo "${APP_HOSTS}" | cut -d" " -f1)
+    hostsArray=("${APP_HOSTS}")
```
ctubbsii, line 67: don't forget to use local variables for anything you don't think should be set outside this function; if assigning a simple value to another local variable, you can do `local x=$y`. For arrays, you can do `local x=("${y[@]}")`. For assigning the result of an operation, you have to declare as local and assign it in two steps: `local x; x=$(echo y)`

My previous suggestion already set APP_HOSTS to an array from the function's input arguments, so that could simplify some of this array logic in here.

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
+    numOfHosts=${#hostsArray[@]}
+    declare -a nodesToWack
+
+    if ((max_kill == 1)) ; then
+      echo "${T} Killing ${app_name}"
+      ssh "${nodeToWack}" "pkill -f '[ ]org.apache.accumulo.start.*${app_name}'"
+    else
+      # get the random nodes to kill
+      count=0
+      while [ $count -lt $numToWack ]; do
```
ctubbsii, line 77: Should use `[[ ]]` instead of `[ ]` to avoid any possible performance hit by calling `/usr/bin/[`, or any possible confusion because it looks like that's what it's doing. `[[ ]]` is built-in, but `[ ]` is a standalone executable that bash may or may not execute or replace with its own built-in interpretation (just like /usr/bin/echo).

bin/agitator
```diff
@@ -30,6 +30,144 @@ Possible commands:
...
+        nodeToWack=${hostsArray[randomHostIndex]}
+        # only add host to the array if its not already there
+        if [[ ! " ${nodesToWack[*]} " =~ ${nodeToWack} ]]; then
+          nodesToWack[count]=${nodeToWack}
+        fi
+        count=${#nodesToWack[@]}
+      done
+      echo "${T} Killing ${count} ${app_name} nodes"
+      for i in "${nodesToWack[@]}"; do
+        ssh "${i}" "pkill -f '[ ]org.apache.accumulo.start.*${app_name}'"
```
ctubbsii, line 88: Could make this pattern be a little more specific (might not be correct):
```suggestion
        ssh "${i}" "pkill -f '[ ]org[.]apache[.]accumulo[.]start[.]Main ${app_name}'"
```

ctubbsii (commented):
Done commenting for now, but left comments that could apply throughout the changes in the script, and not just where I commented.

bin/agitator
milleruntime: The `$APP_HOSTS` variable contains a string of values separated by a space so I was concerned they would expain as separated param if I passed them as a variable. I will test this out.

bin/agitator
milleruntime: What are the benefits of using shift 6 times? Personally I think this is a lot less user friendly then using the numbers for the parameters. If someone comes and looks at the script to figure out the params they have to do counting and also the shifting makes the ordering of setting the variables important.

milleruntime: I see why you are shifting; so you can set the last one as an array. I think I am going to create another function to set APP_HOSTS so this is more readable.

bin/agitator
ctubbsii: They will, but that's okay in the example I gave, because I'm basically passing them as varargs. You can either pass them as a single, quoted value with space separated items, and then put that into an array later:

```bash
function printHosts() {
  local myarray=($1) # make the single argument turn into an array by not quoting so the spaces create separate items
  local x; for x in "${myarray[@]}"; do echo "$x"; done
}
myhosts="host1 host2 host3"
printHosts "$myhosts" # quotes pass it as single arg
```

Or, you can break them out like varargs (which is what I did above and in both the following examples):

Varargs example using an unquoted space-delimited scalar:

```bash
function printHosts() {
  local myarray=("$@") # put args into a new named array
  local x; for x in "${myarray[@]}"; do echo "$x"; done
}
myhosts="host1 host2 host3"
printHosts $myhosts # no quotes, so pass it as 3 args
```

Varargs example using an initial array where each element in the original array is passed as a separate argument:

```bash
function printHosts() {
  local myarray=("$@") # put args into a new named array
  local x; for x in "${myarray[@]}"; do echo "$x"; done
}
myhosts=("host1" "host2" "host3") # already an array
printHosts "${myhosts[@]}" # pass each element of the array as a separate arg
```

(Note: all of these examples are simplified, to avoid the `shift` that picks off the arguments before the varargs hosts.)

milleruntime: Thanks for the tips. I already refactored the script to only use one function for all the agitators (except the hdfs one which still uses perl, that can be a follow on). I found a way to pass a variable nameref to a funtion to set the array for the hosts. Let me know what you think.

milleruntime:
Here is some documentation on the nameref option. I had never used it before, pretty cool.

<pre>
A variable can be assigned the nameref attribute using the -n option to the declare or local builtin commands (see Bash 
Builtins) to create a nameref, or a reference to another variable. This allows variables to be manipulated indirectly. Whenever 
the nameref variable is referenced, assigned to, unset, or has its attributes modified (other than using or changing the nameref 
attribute itself), the operation is actually performed on the variable specified by the nameref variable’s value.
</pre>

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameters.html

ctubbsii:
> Here is some documentation on the nameref option. I had never used it before, pretty cool.

I'm not familiar with the nameref flag. It looks a bit like setting pointers, which seems... harder to grok and more prone to error. Also, bash docs say you can't use it with arrays. I haven't had a chance to look at your updates yet, to see how you're using it.

bin/agitator
```diff
@@ -30,6 +30,81 @@ Possible commands:
...
+function start_app_agitator() {
+  local app_name=$1
+  local kill_sleep_time=$2
+  local restart_sleep_time=$3
+  local min_kill=$4
+  local max_kill=$5
+  local start_cmd=$6
+  local kill_cmd=$7
+  local hosts_array
+  get_app_hosts "$app_name" hosts_array
```
ctubbsii, line 45: The use of a nameref in `get_app_hosts` is clever, but I think it's more complicated than necessary. It's much simpler to `echo` the hosts in `get_app_hosts` and simply capture that output into an array here:

```bash
  local hosts_array; hosts_array=($(get_app_hosts "$app_name"))
```

This would be similar to Java:
```java
String[] hostsArray = getAppHosts(appName).split(" ");
```

Shellcheck won't like not quoting the output, even though that's what you want here. So the best way is to explicitly specify the delimiter when constructing the array, using `mapfile` (aka `readarray`):

```suggestion
  local hosts_array; readarray -td' ' hosts_array < <(get_app_hosts "$app_name")
```

bin/agitator
```diff
@@ -40,57 +115,85 @@ function start_agitator() {
...
+function get_app_hosts() {
+  local app_name=$1
+  local -n app_hosts=$2
+  if [[ "$app_name" == "manager" ]]; then
+    app_hosts=( $MANAGER_HOSTS )
+  fi
+
+  if [[ "$app_name" == "tserver" || "$app_name" == "datanode" ]]; then
+    app_hosts=( $TSERVER_HOSTS )
   fi
```
ctubbsii, line 197: The corresponding change here to my suggestion above could be:

```suggestion
  case "$1" in
    manager)  echo -n "$MANAGER_HOSTS" ;;
    tserver|datanode) echo -n "$TSERVER_HOSTS" ;;
    *) return 1 ;;
  esac
```

The `-n` is so you don't capture a newline as part of the last host when `mapfile` reads it using a space delimiter.

ctubbsii (commented):
Again, didn't have time to go over everything, but hopefully this helps with some incremental improvements.

bin/agitator
```diff
@@ -30,6 +30,81 @@ Possible commands:
...
+  local restart_sleep_time=$3
+  local min_kill=$4
+  local max_kill=$5
+  local start_cmd=$6
+  local kill_cmd=$7
+  local hosts_array
+  get_app_hosts "$app_name" hosts_array
+  local num_hosts=${#hosts_array[@]}
+  local node_to_wack
+  declare -a nodes_to_wack_array
```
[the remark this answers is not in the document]
milleruntime: I think I like `variable=()` better as well. And good catch with the misspelling. I was propagating what was there previously  but there is no reason to keep misspelling it haha.

milleruntime:
> > Here is some documentation on the nameref option. I had never used it before, pretty cool.
> 
> I'm not familiar with the nameref flag. It looks a bit like setting pointers, which seems... harder to grok and more prone to error. Also, bash docs say you can't use it with arrays. I haven't had a chance to look at your updates yet, to see how you're using it.

I didn't see that in the bash docs but with bash there are always more ways to do something wrong than right. If `nameref` shouldn't be used with arrays, then that is interesting because I originally found it in a Stack Overflow post about using it with arrays. https://stackoverflow.com/questions/10582763/how-to-return-an-array-in-bash-without-using-globals

bin/agitator
milleruntime: This is neat. It also seems less obscure or hazardous then using `nameref`. I will use this technique.

I assume the space between the 2 `<` symbols was intentional?

EdColeman:
This is an interesting thread - and I'm asking because I'm curious, not because I disagree with any of the comments or suggestions.  Is there a general style guide / best practices that we can point to?  Right now, if I needed to submit / change a bash script I would be left to find one in our repo and try to match those conventions. Being the scripts might not be uniform because they were written at different points, by different authors and have evolved separately - my odds of picking a "good" example seems less than assured.

bin/agitator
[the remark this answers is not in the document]
milleruntime: Fixed in 2ab7f6e

bin/agitator
```diff
@@ -30,6 +30,80 @@ Possible commands:
...
+        echo "$T Restarting $app_name node at ${i}"
+        ssh "$i" "bash -c '${ENV_VARS} $start_cmd'"
+      done
+    fi
+  done
+}
+
 function start_agitator() {
   ## check that pssh is installed, falling back to parallel-ssh if needed
   ## make sure to export it, so it can be seen inside the agitator perl script
```
DomGarguilo, line 109: Might need to remove/edit this comment to remove reference to the perl script.

bin/agitator
```diff
@@ -30,6 +30,80 @@ Possible commands:
...
+    echo "ERROR: No hosts were found in env for $app_name"
+    exit 1
+  fi
+  if (( max_kill > num_hosts )); then
+      echo "ERROR: Max kill $max_kill greater then number of hosts $num_hosts"
+      exit 1
+    fi
+
+  T="$(date +'%Y%m%d %H:%M:%S')"
+  echo "$T Starting $app_name agitation $kill_sleep_time $restart_sleep_time $min_kill $max_kill for ${hosts_array[*]}"
```
DomGarguilo, line 61: It might be helpful to make this more descriptive. Its hard to tell what the output means without looking at the script. Maybe something to describe what each of the 4 numbers in the message refer to.

milleruntime: I thought about that but I didn't want to make it too long. I'll make it more descriptive.

bin/agitator
milleruntime: Actually, there is still the one perl script that is still being used `hdfs-agitator.pl`. AFAIK its still working. I didn't see any errors but was unable to fully test just locally on my machine.

bin/agitator
milleruntime: Check out 2bf1872 and let me know what you think.

bin/agitator
DomGarguilo: Whoops, I missed that one :+1:

bin/agitator
```diff
@@ -30,6 +30,81 @@ Possible commands:
...
+    exit 1
+  fi
+  if (( max_kill > num_hosts )); then
+      echo "ERROR: Max kill $max_kill greater then number of hosts $num_hosts"
+      exit 1
+    fi
+
+  T="$(date +'%Y%m%d %H:%M:%S')"
+  echo "$T Starting $app_name agitation. Kill every $kill_sleep_time minutes, restart every $restart_sleep_time minutes."
+  echo "$T Will randomly kill $min_kill to $max_kill from the following: ${hosts_array[*]}"
```
DomGarguilo, line 62: Looks good. Minor suggestion that made it a bit more clear to me but either way looks good.
```suggestion
  echo "$T Will randomly kill between $min_kill and $max_kill of the following: ${hosts_array[*]}"
```
An example of this reads: `Will randomly kill between 1 and 1 of the following: localhost`

bin/agitator
```diff
@@ -40,60 +115,84 @@ function start_agitator() {
...
+    sudo -i -u "$AGTR_ACCUMULO_USER"
   fi
-  if [[ $AGITATOR_USER == "$AGTR_HDFS_USER" ]]; then
-    echo "Running datanode-agitator as $AGITATOR_USER"
-    $datanode_cmd > "${datanode_log}.out" 2> "${datanode_log}.err" &
-  else
-    echo "Running datanode-agitator as $AGTR_HDFS_USER using sudo."
-    sudo -u "$AGTR_HDFS_USER" $datanode_cmd > "${datanode_log}.out" 2> "${datanode_log}.err" &
+  echo "Starting manager and tserver agitation as $(whoami)"
+  start_app_agitator manager "$AGTR_MASTER_KILL_SLEEP_TIME" "$AGTR_MASTER_RESTART_SLEEP_TIME" 1 1 "$manager_start_cmd" "$manager_kill_cmd" > "$manager_log" 2>&1 &
```
DomGarguilo, line 139: Are these variables, with "master" in them, `"$AGTR_MASTER_KILL_SLEEP_TIME"` and `"$AGTR_MASTER_RESTART_SLEEP_TIME"` still valid?

milleruntime: Yeah I haven't changed them in `env.sh` because I thought it might be a bigger change to migrate away from master. I was going to do that in a separate change but I can here since I am modifying both files.

bin/agitator
```diff
@@ -30,6 +30,81 @@ Possible commands:
...
+  local node_to_kill
+  nodes_to_kill_array=()
+  local T
+  local ENV_VARS="ACCUMULO_HOME=$ACCUMULO_HOME ZOOKEEPER_HOME=$ZOOKEEPER_HOME HADOOP_HOME=$HADOOP_HOME JAVA_HOME=$JAVA_HOME"
+
+  if (( num_hosts == 0 )); then
+    echo "ERROR: No hosts were found in env for $app_name"
+    exit 1
+  fi
+  if (( max_kill > num_hosts )); then
```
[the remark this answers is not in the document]
milleruntime: Good idea.

bin/agitator
DomGarguilo: > I was going to do that in a separate change but I can here since I am modifying both files.

Either way works. I just wanted to double check that it wasn't being overlooked.

milleruntime:
@DomGarguilo I noticed a quirk while testing that I think could be a follow on task. For the stop agitator, I made the `pkill` kill the sleep first and then anything matching `agitator`. This works but I noticed that the script may run before being killed and introduces a race condition. Here is the log from the tserver agitator:
<pre>
20220112 12:37:19 Starting tserver agitation. Kill every 20 minutes, restart every 10 minutes.
20220112 12:37:19 Will randomly kill between 1 and 1 of the following: localhost
20220112 12:37:19 Sleeping for 20 minutes
Terminated
20220112 12:49:54 Killing tserver at localhost
</pre>

There should be a way to kill the agitator process and all its child processes (i.e. the sleep function) with one command.

DomGarguilo:
> There should be a way to kill the agitator process and all its child processes (i.e. the sleep function) with one command.

Good idea. I can start looking into that if you would like. @milleruntime

PR #189, opened 2022-02-04
Update usages of ThreadPools.createFixedThreadPool

DomGarguilo:
https://github.com/apache/accumulo/pull/2432 introduced a new parameter into `ThreadPools.createFixedThreadPool()`. This PR adds the new parameter to fix the build errors.

The new parameter has been set to false since we most likely wont be looking at any metrics for thread pools created inside the test code.


dlmarion (approved):

Manno15:
:+1:

PR #191, opened 2022-02-17
Make MulitTable reuse tables

milleruntime:
* This allows restarting rw tests and allow using of the same tables
across multiple threads
* Other various improvements to MultiTable

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/multitable/DropTable.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/multitable/DropTable.java
@@ -34,7 +34,7 @@ public class DropTable extends Test {
     List<String> tables = (List<String>) state.get("tableList");
 
     // don't drop a table if we only have one table or less
-    if (tables.size() <= 1) {
+    if (tables.size() <= 25) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/accumulo/testing/randomwalk/multitable/DropTable.java
```diff
@@ -34,7 +34,7 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
     List<String> tables = (List<String>) state.get("tableList");
 
     // don't drop a table if we only have one table or less
```
DomGarguilo, line 36: This comment might need to be updated to reflect the new  if condition.

milleruntime: Fixed in bf9f3fc

milleruntime:
@DomGarguilo think this is OK to merge? I am going to use this for testing.

DomGarguilo (approved):
Looked through the code and ran rwalk with MultiTable.xml. Everything LGTM

PR #200, opened 2022-03-30
Wait for the NameNode to leave safemode before init'ing Accumulo

dlmarion:


dlmarion:
I ran into an issue where Accumulo was being initialized, and failed, because the NameNode was in safemode. The DataNode on the worker machine must not have started up quickly enough. This simple change should prevent that.

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_hadoop.sh.tftpl
```diff
@@ -28,9 +28,15 @@ pdsh -S -g worker sudo systemctl enable hadoop-datanode yarn-nodemanager
...
 # 5. Start datanodes and nodemanagers on the worker nodes
+# 6. Wait for the namenode to leave safemode
 #
 sudo systemctl start zookeeper
 hdfs namenode -format
 sudo systemctl start hadoop-namenode
 sudo systemctl start yarn-resourcemanager
 pdsh -S -g worker 'sudo systemctl start hadoop-datanode yarn-nodemanager'
+echo "Waiting for NameNode to leave safemode..."
+while [[ ! $(hdfs dfsadmin -safemode wait) ]]; do
```
brianloss, line 39: Looking at the [source code](https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/tools/DFSAdmin.java) for this, it would seem that this command waits internally and always returns success. It will print out whether the namenode is actually in safe mode or not, so it would seem that's how you have to check if it failed, but maybe I misread the code.

dlmarion: Are you suggesting that `waitExitSafeMode` does not wait?

https://github.com/apache/hadoop/blob/trunk/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/tools/DFSAdmin.java#L688

brianloss (approved):
I added a suggestion, but otherwise LGTM.

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_hadoop.sh.tftpl
ctubbsii: Alternatively:

```suggestion
echo "Waiting for NameNode to leave safemode..."
while [[ ! $(hdfs dfsadmin -safemode wait 2>/dev/null) =~ OFF ]]; do
  sleep 5
  echo "Waiting for NameNode to leave safemode..."
done
```

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_hadoop.sh.tftpl
```diff
@@ -28,9 +28,24 @@ pdsh -S -g worker sudo systemctl enable hadoop-datanode yarn-nodemanager
...
+while [[ $exit_code -ne 0 ]]; do
+  sleep 5s
+  echo "Waiting for NameNode to leave safemode..."
+  hdfs dfsadmin -safemode wait > /tmp/namenode_safemode_status
+  exit_code=$?
+done
+if ! [[ $(cat /tmp/namenode_safemode_status) =~ OFF ]]; then
+  echo "Finished waiting for SafeMode to exit, but safemode is not OFF"
+fi
+rm -f /tmp/namenode_safemode_status
```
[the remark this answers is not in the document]
ctubbsii: Alternatively:

```suggestion
echo "Waiting for NameNode to leave safemode..."
while [[ ! $(hdfs dfsadmin -safemode wait 2>/dev/null) =~ OFF ]]; do
  sleep 5
  echo "Waiting for NameNode to leave safemode..."
done
```

brianloss (approved):

PR #201, opened 2022-04-07
Update OpenTelemetry and Micrometer versions to match Accumulo 2.1.0

dlmarion:


ctubbsii (approved):
For something this trivial, to sync it with what's already in 2.1.0, especially for the testing repo, and also especially when there's no other changes than version bumps, you probably should just commit it. Not much to even bother to review here. :smiley_cat:

PR #206, opened 2022-05-10
runs mvn install for accumulo in terraform

keith-turner:
Run mvn install when building accumulo so that the accumulo-testing will pick up the correct accumulo snapshot jars.


brianloss (approved):
LGTM

dlmarion (approved):

ctubbsii (approved):
One of the rare cases where install actually makes sense over package. :smiley_cat:

PR #207, opened 2022-05-10
Terraform config fixes based on testing

brianloss:
* Update AzureRM Terraform provider to use any 3.x version.
* Fix bug where variable validation would fail for the Azure
  configuration if managed_disk_configuration was not used and set to
  null.
* Clarify in the README that terraform init/plan/apply commands are
  sensitive to the current working directory.
* Clarify in the README that the appropriate SSH private key must be
  present in the SSH agent in order for terraform apply to succeed.


dlmarion (approved):
LGTM

dlmarion (approved):

PR #208, opened 2022-05-11
Create BadFile RW Test

milleruntime:
@EdColeman This was how I created an RFile to try and simulate a real data scenario. Hopefully this will allow us to do more then just have a file that isn't closed.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/State.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/State.java
@@ -181,4 +181,17 @@ public class State {
   public Long getLong(String key) {
     return (Long) stateMap.get(key);
   }
+
+  /**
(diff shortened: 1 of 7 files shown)


DomGarguilo:
Is this intended to be merged into main or is this just meant to serve as an example? @milleruntime 

I looked over thing briefly and it looks good to me, just wanted to check before looking into this further.

milleruntime:
From what I remember, it works but pretty much just kills the test. I am thinking that it might be better to have it as a separate test utility. @DomGarguilo if you are interested, feel free to take over and run with it. I am pretty busy with other tasks at the moment.

ctubbsii:
There are conflicts with the main branch now. I'm going to close this for now, as it is stale. But it can be re-opened if somebody wants to pick it back up.

PR #212, opened 2022-07-25
Miscellaneous improvements to continuous code

DomGarguilo:
These changes include:
* Add a `Random` member variable to `TestEnv` to be used where possible instead of constructing new ones when needed.
* Add more `AutoClosable`s to try-with-resources blocks
* Simplify some logic using streams
* Improve some log messages

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousBatchWalker.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousBatchWalker.java
@@ -18,7 +18,7 @@ package org.apache.accumulo.testing.continuous;
 
 import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
 
-import java.util.ArrayList;
+import java.time.Duration;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -26,6 +26,7 @@ import java.util.Map.Entry;
 import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.client.BatchScanner;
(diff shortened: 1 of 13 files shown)


src/main/java/org/apache/accumulo/testing/continuous/ContinuousBatchWalker.java
```diff
@@ -39,34 +40,33 @@
...
-      while (true) {
-        BatchScanner bs = client.createBatchScanner(env.getAccumuloTableName(), auths);
 
-        Set<Text> batch = getBatch(scanner, env.getRowMin(), env.getRowMax(), scanBatchSize, r);
-        List<Range> ranges = new ArrayList<>(batch.size());
+      Authorizations auths = env.getRandomAuthorizations();
 
-        for (Text row : batch) {
-          ranges.add(new Range(row));
+      try (AccumuloClient client = env.getAccumuloClient();
```
milleruntime, line 54: Does the client get reused? I am wondering since it gets pulled from the `env`. I don't think it does, its just a convenience method...

milleruntime: OK its probably not a big deal then but I'd remove the try-with-resources around the client since its redundant.

milleruntime, line 54: Client doesn't need to be in the try-with-resources.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousScanner.java
```diff
@@ -40,58 +34,51 @@ public class ContinuousScanner {
 
   public static void main(String[] args) throws Exception {
 
-    try (ContinuousEnv env = new ContinuousEnv(args)) {
+    try (ContinuousEnv env = new ContinuousEnv(args);
+        AccumuloClient client = env.getAccumuloClient()) {
```
milleruntime, line 38: Same here.

milleruntime (approved):
Looks good overall. Only thing I had was that the closing of client in the try-with-resources could cause errors.

src/main/java/org/apache/accumulo/testing/continuous/ContinuousBatchWalker.java
DomGarguilo: Addressed these in [8a658d6](https://github.com/apache/accumulo-testing/pull/212/commits/8a658d64fe381d511292c24145cd46602d5bf66f)

PR #221, opened 2022-09-14
Drop security from all modules

milleruntime:
* Also sort modules printed by script

### DIFF
--- a/bin/rwalk
+++ b/bin/rwalk
@@ -28,7 +28,7 @@ Usage: $0 <module> (<argument>)
    Modules listed below located in <at_home>/src/main/resources/randomwalk/modules
 
 EOF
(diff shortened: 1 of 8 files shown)


DomGarguilo:
This is looking good. I was testing the modules though and ran into an error with running all of these modules on their own. I don't think its related to these change though as it also happens on main so ill open a separate ticket for that.

DomGarguilo:
> This is looking good. I was testing the modules though and ran into an error with running all of these modules on their own. I don't think its related to these change though as it also happens on main so ill open a separate ticket for that.

Nevermind. Not sure what was going on there but after rebuilding and restarting uno, everything is working fine.

DomGarguilo (approved):

PR #222, opened 2022-09-16
Upgrade Maven, ZooKeeper, and Hadoop versions

dlmarion:


EdColeman (approved):

DomGarguilo:
I noticed there is a newer version of the aws provider that might be able to be updated. I don't know if this makes much of a difference or will cause any issues but I thought I would point it out just in case it would be worthwhile to include into this PR.

According to https://registry.terraform.io/providers/hashicorp/aws/latest it looks like its up to version 4.31

https://github.com/apache/accumulo-testing/blob/e8e69a054f46979cec499724e6931bfb77947f23/contrib/terraform-testing-infrastructure/shared_state/aws/main.tf#L22

https://github.com/apache/accumulo-testing/blob/e9cbb856a0fac3335fff13fc175d7439c0940cca/contrib/terraform-testing-infrastructure/aws/main.tf#L62

dlmarion:
@DomGarguilo - yeah, good thought. I will update that in this branch on Monday and test it out.

ctubbsii (approved):

ctubbsii:
Hadoop has 3.3.4 now, but this is good with or without that update.

Manno15 (approved):

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_accumulo.sh.tftpl
```diff
@@ -22,7 +22,7 @@ docker run -d --name jaeger \
     -p 6832:6832/udp -p 5778:5778 \
     -p 16686:16686 -p 14268:14268 \
     -p 14250:14250 -p 9411:9411 \
-    jaegertracing/all-in-one:1.29
+    artifactory.adv.evoforge.org/docker-remote/jaegertracing/all-in-one:1.29
```
brianloss, line 25: I don't believe this is a publicly accessible image repository. At least, I'm not able to pull from it.
```bash
$ docker pull artifactory.adv.evoforge.org/docker-remote/jaegertracing/all-in-one:1.29
Error response from daemon: Get "https://artifactory.adv.evoforge.org/v2/": Failed to lookup host: artifactory.adv.evoforge.org
```

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_accumulo.sh.tftpl
```diff
@@ -36,7 +36,7 @@ docker run --ulimit nofile=66000:66000 -d \
     -v /data/metrics/grafana:/var/lib/grafana \
     -v ${software_root}/telegraf/conf:/etc/telegraf \
     -v ${software_root}/grafana/dashboards:/etc/grafana/provisioning/dashboards \
-    artlov/docker-telegraf-influxdb-grafana:latest
+    artifactory.adv.evoforge.org/docker-remote/artlov/docker-telegraf-influxdb-grafana:latest
```
brianloss, line 39: Same issue here...

brianloss (changes requested):

contrib/terraform-testing-infrastructure/modules/config-files/templates/initialize_accumulo.sh.tftpl
dlmarion: Thx, side effect of `git commit -a`.

brianloss: You could push a repo name through as a variable (default it to docker.io) if you want to be able to change where these images are pulled from w/o having to modify the source code.

brianloss (approved):

dlmarion:
Pushing a commit here in a minute with changes from testing. The upgrade of the aws provider from 3.68 to 4.31 deprecated some of the S3 configuration with a new way to specify it. I removed the deprecated items and replaced them with the new way of doing things in c872d9e. However, in testing, the new S3 ACL resource was returning a 403. Backing out just that change allowed things to work. Also note that if you had used version 3.68 in the past, then this will require you to perform an update. There's a nice error message with instructions on the command you need to run.

PR #235, opened 2022-10-05
fixes two bulk import RW bugs

keith-turner:
One bug is an off by one bug introduced when changing code from a loop to a stream. After that change the set was one larger than it used to be.  This caused the test to generate less data than intended for each import which inevitably caused the counts to be non-zero. 

The other bug is using a scanner outside of a try with resources block that closes the scanner.


keith-turner:
Ran into these while working on trying to reproduce apache/accumulo#2667

EdColeman (approved):

DomGarguilo (approved):

PR #236, opened 2022-10-06
fixes bug with bulk RW file partition point creation

keith-turner:
There was code that did the following

    TreeSet<Integer> startRows = new TreeSet<>();
    startRows.add(0);
    while (startRows.size() < parts)
      startRows.add(rand.nextInt(LOTS));

The above code was replaced in 7453c37 with a stream. The stream did not fully capture the original behavior of the loop.  This change makes the stream fully capture that behavior.  Need to ensure that `parts` unique random numbers are generated including zero (like if the random number generator returns zero it should be properly deduplicated).  The stream was not properly handling the RNG returning duplicates or zero.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/BulkPlusOne.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/BulkPlusOne.java
@@ -59,9 +59,9 @@ public class BulkPlusOne extends BulkImportTest {
     log.debug("Bulk loading from {}", dir);
     final int parts = env.getRandom().nextInt(10) + 1;
 
-    TreeSet<Integer> startRows = Stream.generate(() -> env.getRandom().nextInt(LOTS))
-        .limit(parts - 1).collect(Collectors.toCollection(TreeSet::new));
-    startRows.add(0);
+    TreeSet<Integer> startRows = Stream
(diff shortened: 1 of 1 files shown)


keith-turner:
I'm beginning to wonder if we should replace the stream with the original loop.  The created set must always have at least `0` in it.  Will the stream changes I created with Stream.concat result in this?

keith-turner:
> I'm beginning to wonder if we should replace the stream with the original loop. The created set must always have at least 0 in it. Will the stream changes I created with Stream.concat result in this?

I looked at the stream javadoc and it will maintain order, so since `0` comes first it should always be there.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#concat(java.util.stream.Stream,java.util.stream.Stream)

I think the stream code may be slightly more readable than the loop with the explicit call to distinct(). So I think its slightly better than the loop.  I am going to add a comment about zero.

keith-turner:
I lost the exception, but I was seeing error when run the Bulk RW test.  When duplicate rand nums were generated it would create a set that was too small and this would result in an index out bounds exception on a later loop.  This PR is in support of running bulk RW test for apache/accumulo#2667.

DomGarguilo (approved):

EdColeman (approved):

PR #238, opened 2022-10-06
Use read -a instead of readarray -td' '

### LINKED ISSUE #237: agitator fails on CentOS 7.9
Error in the agitator logs:

```
./agitator: line 44: readarray: -d: invalid option
readarray: usage: readarray [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
ERROR: No hosts were found in env for tserver
```

Running with bash `GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)` on `CentOS Linux release 7.9.2009 (Core)`

ctubbsii:
Use built-in read command to do word splitting naturally, instead of relying on readarray/mapfile's newer delimiter flag to do word splitting forcibly.

Also fix some other cases where read -a would be better.

This fixes #237

### DIFF
--- a/bin/agitator
+++ b/bin/agitator
@@ -41,7 +41,7 @@ function start_app_agitator() {
   local max_kill=$5
   local start_cmd=$6
   local kill_cmd=$7
-  local hosts_array; readarray -td' ' hosts_array < <(get_app_hosts "$app_name")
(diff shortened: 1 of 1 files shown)


dlmarion (approved):
Tested on CentOS7, appears to work

PR #245, opened 2022-10-14
Standardize formatting of shell scripts

ctubbsii:
* Standardize shebang lines for bash scripts
* Run `shfmt -ln bash -l -d -i 2 -ci -s .` to format scripts

### DIFF
--- a/bin/agitator
+++ b/bin/agitator
@@ -18,8 +18,8 @@
 # under the License.
 #
 
-bin_dir=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
-at_home=$( cd "$( dirname "$bin_dir" )" && pwd )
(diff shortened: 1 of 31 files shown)


DomGarguilo:
These changes look good.

> * Standardize shebang lines for bash scripts

I also noticed there are files that have `#! /bin/bash` instead of `#! /usr/bin/env bash`. Not sure if this is something that should be standardized or if it should be handled here but just wanted to point it out.

ctubbsii:
> These changes look good.
> 
> > * Standardize shebang lines for bash scripts
> 
> I also noticed there are files that have `#! /bin/bash` instead of `#! /usr/bin/env bash`. Not sure if this is something that should be standardized or if it should be handled here but just wanted to point it out.

I thought I fixed all those, but it may have missed the terraform template ones.

Issue #246, opened 2022-10-14
Investigate pulling OpenTelemetry and Micrometer version from Accumulo repo


dlmarion:
When OpenTelemetry and Micrometer versions change in the accumulo repo, they need to be updated in the terraform configuration directory to keep them in sync (ref [commit](https://github.com/apache/accumulo-testing/commit/049163a8a20623a7d240bc65efe92738edf6f301). Investigate making this automatic. We might be able to remove the specific version references from the README and pull the versions from the pom and use them in the template somehow.

ctubbsii:
I did this same thing in apache/accumulo-examples#84

PR #247, opened 2022-10-14
Update OpenTelemetery and Micrometer versions with Accumulo latest

cshannon:
This will be needed after merging https://github.com/apache/accumulo/pull/3022


dlmarion (approved):

EdColeman:
With https://github.com/apache/accumulo-testing/issues/246 this may not be necessary - no issue with merging so that versions remain consistent - but it would really be nice if we didn't need to remember that these projects need to remain synchronized.

cshannon:
I will merge this for now as #246 hasn't been done yet but maybe in the future we won't need to do this

PR #251, opened 2022-10-24
Enable reloading log4j2 config

keith-turner:
While running random walk test one of the walkers got stuck.  I tried to change the accumulo testing log4j2 config to enable accumulo client debugging but nothing happened.  Looking into it I think an additional setting is needed in the log4j config for this to work.

### DIFF
--- a/conf/log4j2.properties
+++ b/conf/log4j2.properties
@@ -20,6 +20,7 @@
 name = AccumuloTestingDefaultLoggingProperties
 status = info
 dest = err
+monitorInterval = 30
 
 appenders = console
 appender.console.type = Console


dlmarion (approved):

ctubbsii (approved):

keith-turner:
@ctubbsii for my processes that were started w/o this config option would you happen to know of any way to force them to reload log4j2 config?  I have been looking around hoping to find something, but so far I have not found anything.

ctubbsii:
> would you happen to know of any way to force them to reload log4j2 config?

No, sorry. I don't think it's possible. My understanding is that it doesn't monitor for changes unless that option is set up front.

PR #252, opened 2022-10-27
Only install & enable Docker on manager node

dlmarion:


dlmarion:
The reason for this change is that `cloud-init` would intermittently fail with:
```
(remote-exec): ('package-update-upgrade-install', ProcessExecutionError(u"Unexpected error while running command.\nCommand: ['yum', '-t', '-y', 'makecache']\nExit code: 1\nReason: -\nStdout: -\nStderr: -",))
```

It would routinely take 3 or 4 attempts to create a cluster. The `cloud-init` log file contained:
```
2022-10-27 11:53:30,436 - helpers.py[DEBUG]: Running update-sources using lock (<FileLock using file '/var/lib/cloud/instances/i-0ab76c54cf53177d8/sem/update_sources'>)
2022-10-27 11:53:30,437 - rhel.py[DEBUG]: Using YUM for package management
2022-10-27 11:53:30,437 - util.py[DEBUG]: Running command ['yum', '-t', '-y', 'makecache'] with allowed return codes [0] (shell=False, capture=False)
2022-10-27 11:53:31,964 - util.py[WARNING]: Package update failed
2022-10-27 11:53:31,965 - util.py[DEBUG]: Package update failed
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/cloudinit/config/cc_package_update_upgrade_install.py", line 85, in handle
    cloud.distro.update_package_sources()
  File "/usr/lib/python2.7/site-packages/cloudinit/distros/rhel.py", line 180, in update_package_sources
    ["makecache"], freq=PER_INSTANCE)
  File "/usr/lib/python2.7/site-packages/cloudinit/helpers.py", line 187, in run
    results = functor(*args)
  File "/usr/lib/python2.7/site-packages/cloudinit/distros/rhel.py", line 176, in package_command
    util.subp(cmd, capture=False)
  File "/usr/lib/python2.7/site-packages/cloudinit/util.py", line 2084, in subp
    cmd=args)
ProcessExecutionError: Unexpected error while running command.
Command: ['yum', '-t', '-y', 'makecache']
Exit code: 1
Reason: -
Stdout: -
Stderr: -
```

And `/var/log/messages` contained:
```
cloud-init: Error making cache directory: /var/cache/yum/x86_64/7/docker error was: [Errno 17] File exists: '/var/cache/yum/x86_64/7/docker'
```

From what I could tell `cloud-init` was running at the same time as a `salt-minion` in my environment. I think the `salt-minion` (or the AMI) is conflicting with `cloud-init` sometimes, probably a race condition. This change installs the Docker repo on CentOS, but disables it, such that the `yum makecache` command that is run by the `cloud-init` process does not try to create the directory. Docker is only needed on the `manager` node, so I then install and enable it in the initialize accumulo script before the Docker containers are started.

brianloss:
> Docker is only needed on the manager node, so I then install and enable it in the initialize accumulo script before the Docker containers are started.

I recently did a test on rc0 where I enabled swarm and used docker on the whole cluster. Plus, isn’t this going to happen eventually even on the manager if it’s a race condition with salt and cloud-init?

I’m not strongly opposed to this (esp since I don’t have a better idea of what to change to fix right now), but it kind of feels to me like this change may be pushing some tech debt that we’ll have to deal with later.

dlmarion:
> Plus, isn’t this going to happen eventually even on the manager if it’s a race condition with salt and cloud-init?

I'm not sure, but if it does happen we might have a better way of dealing with it being here rather than in cloud-init. The `yum makecache` command is buried in the `cloud-init` code with no way for us to deal with errors. If we want Docker on all nodes, we could push the installation and setup of Docker to the end of the `install_sw.sh.tftpl` file.

brianloss (approved):
Good enough for now. I do still worry that there will be other issues if two competing systems to configure the machines are running.

dlmarion:
> I do still worry that there will be other issues if two competing systems to configure the machines are running.

Agreed. Nothing I can do about that currently in my environment. I'm going to bring it up though and see what can be done.

PR #254, opened 2022-11-05
Refactor logging in Verify for RW Bulk

### LINKED ISSUE #239: Make bulk import random walk verify step print missing markers
While debugging apache/accumulo#3006 I spent a while trying to determine what markers were missing.   It would be nice to have the verify code just print this information out when it fails.  In the test that failed the following error was printed

```
Caused by: java.lang.Exception: Bad key at r00000 cf:000 [] 1665128566432 false=-1
```

Through manual analysis I determined that marker 153 was missing.  Would be nice if the error message gave this info, would really speed up debugging issues found by the test.  Thinking maybe an error message like the following would be really useful.

```
Caused by: java.lang.Exception: Bad key at r00000 cf:000 [] 1665128566432 false=-1 missing markers : [0000153]
```

The code that prints this error message is here

https://github.com/apache/accumulo-testing/blob/ddc6e589a6703239b7445075172fe88308e5a0ef/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java#L73

Below that code is some code that analyzes markers, but it does not run when the count is not zero.

Manno15:
Closes #239.

 Instead of throwing an exception at each possible failure, an error log is recorded instead. Currently, have it to break out of each loop but we could instead complete the loops and log each possible error. Need to do more testing there to see if that would be worthwhile.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
@@ -67,7 +67,7 @@ public class Verify extends Test {
     String user = env.getAccumuloClient().whoami();
     Authorizations auths = env.getAccumuloClient().securityOperations().getUserAuthorizations(user);
     RowIterator rowIter;
-    Boolean errorFound = false;
+    boolean errorFound = false;
     try (Scanner scanner = env.getAccumuloClient().createScanner(Setup.getTableName(), auths)) {
       scanner.fetchColumnFamily(BulkPlusOne.CHECK_COLUMN_FAMILY);
       for (Entry<Key,Value> entry : scanner) {
@@ -99,12 +99,14 @@ public class Verify extends Test {
           if (curr - 1 != prev) {
             log.error("Bad market count. Current row: {} {}, Previous row marker: {}",
                 entry.getKey(), entry.getValue(), prev);
+            errorFound = true;
             break;
           }
 
           if (!entry.getValue().toString().equals("1")) {
             log.error("Bad marker value for row {} {}.\n Value expected to be one", entry.getKey(),
                 entry.getValue());
+            errorFound = true;
             break;
           }
 
@@ -114,6 +116,7 @@ public class Verify extends Test {
         if (BulkPlusOne.counter.get() != prev) {
           log.error("Row {} does not have all markers. Current marker: {}, Previous marker:{}",
               rowText, BulkPlusOne.counter.get(), prev);
+          errorFound = true;
           break;
         }
       }
--- a/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
+++ b/src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
@@ -75,7 +75,6 @@ public class Verify extends Test {
         if (!Arrays.equals(value, zero)) {
           log.error("Bad key found at {}", entry);
           errorFound = true;
-          break;
         }
       }
 
@@ -100,14 +99,12 @@ public class Verify extends Test {
             log.error("Bad market count. Current row: {} {}, Previous row marker: {}",
                 entry.getKey(), entry.getValue(), prev);
             errorFound = true;
-            break;
           }
 
           if (!entry.getValue().toString().equals("1")) {
             log.error("Bad marker value for row {} {}.\n Value expected to be one", entry.getKey(),
                 entry.getValue());
             errorFound = true;
-            break;
           }
 
           prev = curr;
@@ -117,7 +114,6 @@ public class Verify extends Test {
           log.error("Row {} does not have all markers. Current marker: {}, Previous marker:{}",
               rowText, BulkPlusOne.counter.get(), prev);
           errorFound = true;
-          break;
         }
       }
       if (errorFound) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
```diff
@@ -67,12 +67,15 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
     String user = env.getAccumuloClient().whoami();
     Authorizations auths = env.getAccumuloClient().securityOperations().getUserAuthorizations(user);
     RowIterator rowIter;
+    Boolean errorFound = false;
```
ctubbsii, line 70:
```suggestion
    boolean errorFound = false;
```

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
```diff
@@ -87,26 +90,36 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
...
 
           long curr = Long.parseLong(entry.getKey().getColumnQualifier().toString());
 
-          if (curr - 1 != prev)
-            throw new Exception(
-                "Bad marker count " + entry.getKey() + " " + entry.getValue() + " " + prev);
+          if (curr - 1 != prev) {
+            log.error("Bad market count. Current row: {} {}, Previous row marker: {}",
+                entry.getKey(), entry.getValue(), prev);
+            break;
```
keith-turner, line 102: Will breaking stop it from printing other problems in the test data?  

Should this set errorFound?

```suggestion
            errorFound=true;
            break;
```

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
```diff
@@ -87,26 +90,36 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
...
+          }
 
-          if (!entry.getValue().toString().equals("1"))
-            throw new Exception("Bad marker value " + entry.getKey() + " " + entry.getValue());
+          if (!entry.getValue().toString().equals("1")) {
+            log.error("Bad marker value for row {} {}.\n Value expected to be one", entry.getKey(),
+                entry.getValue());
+            break;
```
keith-turner, line 108:
```suggestion
            errorFound=true;
            break;
```

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
```diff
@@ -87,26 +90,36 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
...
 
           prev = curr;
         }
 
         if (BulkPlusOne.counter.get() != prev) {
-          throw new Exception("Row " + rowText + " does not have all markers "
-              + BulkPlusOne.counter.get() + " " + prev);
+          log.error("Row {} does not have all markers. Current marker: {}, Previous marker:{}",
+              rowText, BulkPlusOne.counter.get(), prev);
+          break;
```
keith-turner, line 117:
```suggestion
          errorFound=true;
          break;
```

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
Manno15: Whoops. Will fix.

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
Manno15: I forgot to add these. and the ones below but yes, they should. I will do more testing on the breaks, but it should match the same behavior it previously did there. As I mentioned in the description, we could forgo the breaks and then log everything if that is the desired behavior.

Manno15:
https://github.com/apache/accumulo-testing/pull/254/commits/f5ff88e2130e57be61d7a4a25f54f4dbe7c3b064 addresses the comments.

src/main/java/org/apache/accumulo/testing/randomwalk/bulk/Verify.java
```diff
@@ -87,26 +90,39 @@ public void visit(State state, RandWalkEnv env, Properties props) throws Excepti
...
           long curr = Long.parseLong(entry.getKey().getColumnQualifier().toString());
 
-          if (curr - 1 != prev)
-            throw new Exception(
-                "Bad marker count " + entry.getKey() + " " + entry.getValue() + " " + prev);
+          if (curr - 1 != prev) {
+            log.error("Bad market count. Current row: {} {}, Previous row marker: {}",
+                entry.getKey(), entry.getValue(), prev);
+            errorFound = true;
+            break;
```
Manno15, line 103: As it stands, this break will be hit and break out of the inner while loop and avoid both of the next checks since `prev` will never be reassigned value to `curr` which is at the end of the inner while loop. Removing both `breaks` in the inner loop would hit both checks but unsure of many redundant errors would get logged, if any.

Manno15: @keith-turner  Any thoughts on this and what approach I should take?

keith-turner: > @keith-turner Any thoughts on this and what approach I should take?

Removing all breaks seems like a good way to go, as long as what it logs is accurate I think its ok if it logs a bunch.  Its good to see all of the problems in the data.

>  It could also affect the outer if statment since prev is never assigned the value of curr which happens at the end of the inner while loop

I agree the breaks in the inner loop could cause the check following the inner loop to report a false positive.

Manno15: I will make these changes

Manno15:
Tests are failing due to the version change of accumulo. I will look into fixing it tomorrow. This PR should be ready for review still.

keith-turner (approved):

PR #256, opened 2022-11-19
Update pom to use accumulo 3.0.0-SNAPSHOT instead of 2.1.0

Manno15:
This also fixes `ShardTool` which broke due to this change. This should fix the test failures seen in accumulo-testing. I am unsure on if we plan to keep `ShardTool` since it was still using a deprecated class that 3.0.0 aims to remove.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
   <name>Apache Accumulo Testing</name>
   <description>Testing tools for Apache Accumulo</description>
   <properties>
-    <accumulo.version>2.1.0-SNAPSHOT</accumulo.version>
+    <accumulo.version>3.0.0-SNAPSHOT</accumulo.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -32,7 +32,7 @@
   <name>Apache Accumulo Testing</name>
   <description>Testing tools for Apache Accumulo</description>
   <properties>
-    <accumulo.version>2.1.0-SNAPSHOT</accumulo.version>
```
keith-turner, line 35: I wonder if changing this to 2.1.1-SNAPSHOT would have also fixed this, maybe those are available on the apache snapshot repo, not sure.

keith-turner (approved):

pom.xml
Manno15: It is possible, but this matches what is in the main accumulo repo.

PR #257, opened 2022-11-19
fixes build issue by supplying Accumulo version

keith-turner:
Tried to run bin/build against accumulo 2.1.1-SNAPSHOT and it failed looking for 2.1.0-SNAPSHOT resources.  This was because script was not passing the runtime Accumulo version.  Modified the script to pass the runtime version.


Manno15:
I fixed this in #256

Manno15 (approved):

Manno15:
The QA check should be fixed with the changes in #256

keith-turner:
I think #256 would have hidden this problem, so I got lucky in that I attempted to use this repo before #256 was merged.  The build script was not always using what was maven installed locally.  Only uncovered this problem because 2.1.0-SNAPSHOT is no longer in the apache snapshot repo.

keith-turner:
I'll merge this after #256, then maybe we can see if this one build.

Manno15:
I re-ran the build.

PR #264, opened 2023-06-28
Bump maven to 3.8.8 in terraform config

### LINKED ISSUE #263: Terraform setup fails due to old maven version
When running the terraform setup, it got to the step where maven was attempted to be downloaded. A 404 was thrown since the current version default "3.8.6" does not have a binary at the link it was looking for anymore.

Suggested fix:
update the default maven version to pull "3.8.8"

```
module.configure_nodes.null_resource.configure_manager_node (remote-exec): --2023-06-28 16:05:22--  https://dlcdn.apache.org/maven/maven-3/3.8.6/binaries/apache-maven-3.8.6-bin.tar.gz
module.configure_nodes.null_resource.configure_manager_node (remote-exec): Resolving [dlcdn.apache.org](http://dlcdn.apache.org/) ([dlcdn.apache.org](http://dlcdn.apache.org/))... 151.101.2.132, 2a04:4e42::644
module.configure_nodes.null_resource.configure_manager_node (remote-exec): Connecting to [dlcdn.apache.org](http://dlcdn.apache.org/) ([dlcdn.apache.org](http://dlcdn.apache.org/))|151.101.2.132|:443... connected.
module.configure_nodes.null_resource.configure_manager_node (remote-exec): HTTP request sent, awaiting response... 404 Not Found
module.configure_nodes.null_resource.configure_manager_node (remote-exec): 2023-06-28 16:05:22 ERROR 404: Not Found.
```

DomGarguilo:
Fixes #263

### DIFF
--- a/contrib/terraform-testing-infrastructure/README.md
+++ b/contrib/terraform-testing-infrastructure/README.md
@@ -169,7 +169,7 @@ The table below lists the variables and their default values that are used in th
 | instance\_count | The number of EC2 instances to create | `string` | `"2"` | no |
 | instance\_type | The type of EC2 instances to create | `string` | `"m5.2xlarge"` | no |
 | local\_sources\_dir | Directory on local machine that contains Maven, ZooKeeper or Hadoop binary distributions or Accumulo source tarball | `string` | `""` | no |
-| maven\_version | The version of Maven to download and install | `string` | `"3.8.6"` | no |
+| maven\_version | The version of Maven to download and install | `string` | `"3.8.8"` | no |
 | optional\_cloudinit\_config | An optional config block for the cloud-init script. If you set this, you should consider setting cloudinit\_merge\_type to handle merging with the default script as you need. | `string` | `null` | no |
 | private\_network | Indicates whether or not the user is on a private network and access to hosts should be through the private IP addresses rather than public ones. | `bool` | `false` | no |
 | root\_volume\_gb | The size, in GB, of the EC2 instance root volume | `string` | `"300"` | no |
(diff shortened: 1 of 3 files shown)


ctubbsii (approved):
This is fine, but I wonder if it would be better to just have the instance do a `yum install` or `apt-get install` the version of Maven that ships with the OS of the base image, so we don't have to keep updating this. Maven is pretty stable from version to version, so it really doesn't matter which version we use, as long as it's relatively recent. We define the plugin versions we care about ourselves.

DomGarguilo:
> This is fine, but I wonder if it would be better to just have the instance do a `yum install` or `apt-get install` the version of Maven that ships with the OS of the base image, so we don't have to keep updating this. Maven is pretty stable from version to version, so it really doesn't matter which version we use, as long as it's relatively recent. We define the plugin versions we care about ourselves.

Seems like a good idea to me. I wonder if that would make it any harder to install a specific version of maven. I guess we might want to still have it as an option to pin a certain version for testing but again, not sure.

DomGarguilo:
Going to merge this for now. Making the change mentioned above can be handled in a separate ticket.

dlmarion:
> This is fine, but I wonder if it would be better to just have the instance do a yum install or apt-get install the version of Maven that ships with the OS of the base image, so we don't have to keep updating this.

I think the only issue with this is that I think the Maven version for CentOS7 is really old, and CentOS7 is still widely used. I think it might be version 3.0.5.

ctubbsii:
I don't think we need to still be supporting CentOS 7 for our testing repo's main branch (accumulo &gt;= 3.0). The terraform scripts can be updated to use something newer, if they aren't already.

PR #265, opened 2023-09-08
Bump parent pom and accumulo versions

DomGarguilo:
Bumps the parent pom to version 30 to match the version in the accumulo pom and bumps the accumulo version to 3.1.0-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -277,7 +277,7 @@
           <dependency>
             <groupId>com.puppycrawl.tools</groupId>
             <artifactId>checkstyle</artifactId>
-            <version>9.1</version>
+            <version>10.12.3</version>
           </dependency>
         </dependencies>
         <executions>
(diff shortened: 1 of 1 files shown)


EdColeman (approved):

Manno15 (approved):

pom.xml
```diff
@@ -24,15 +24,15 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>27</version>
+    <version>30</version>
```
ctubbsii, line 27: This is fine on its own probably, but I'd bet there are some plugin versions that can be updated at the same time, or whose versions are no longer needed to be specified as of the latest parent POM.

DomGarguilo: I updated the puppycrawl version but that is the only additional change I could find given your suggestions.

ctubbsii: I didn't see any versions that could be removed, but I did bump some plugin versions and made some other follow-on improvements in 9cdafdb8c07d4e815b072aca70cf2ec0acc2df0c

PR #270, opened 2024-01-03
Updates to Terraform configuration based on recent changes

dlmarion:
This commit bumps the OpenTelemetry and Micrometer versions to match the versions in the current 2.1 branch pom.xml file. This also adds variables for os distribution and version to that special handling could be added to install both Java 11 and Java 17 (recent changes to Accumulo require Java 17 to build). Accumulo still uses Java 11 at runtime.


brianloss (approved):
Haven't tried to run it myself, but changes LGTM.

PR #271, opened 2024-01-09
Fix compaction service cs1

ddanielr:
Updates the compaction service details to use the new property and correct format.

### DIFF
--- a/conf/accumulo-testing.properties
+++ b/conf/accumulo-testing.properties
@@ -46,10 +46,10 @@ test.ci.common.auths=
 # Accumulo tserver properties to set when creating a table
 test.ci.common.accumulo.server.props=\
 compaction.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
-compaction.service.cs1.planner.opts.executors=\
-[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
-{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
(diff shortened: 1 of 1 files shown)


ddanielr:
QA verify build step is failing trying to get an accumulo-4.0.0-snapshot from the apache maven repo. 
manually merged this change to bypass the QA build step

conf/accumulo-testing.properties
```diff
@@ -46,10 +46,10 @@ test.ci.common.auths=
...
 test.ci.common.accumulo.server.props=\
 compaction.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
-compaction.service.cs1.planner.opts.executors=\
-[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
-{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
-{"name":"large","type":"internal","numThreads":2}]
+compaction.service.cs1.planner.opts.groups=\
+[{"name":"small","maxSize":"16M"},\
+{"name":"medium","maxSize":"128M"},\
+{"name":"large"}]
```
dlmarion, line 52: Will these names work? They don't match the names in the compactor config at https://github.com/apache/accumulo-testing/blob/elasticity/contrib/terraform-testing-infrastructure/modules/config-files/templates/cluster.yaml.tftpl#L41

ddanielr: No you're right. They will need to be modified to match the resource groups that are created.

PR #276, opened 2024-04-12
Add zipfian distribution option to vary value size for continuous ingest

DomGarguilo:
This PR adds an optional component to the value created in continuous ingest. A random portion of data will be inserted into the value whose size is determined via a zipfian distribution.

The motivation behind this is to add optional variance to the sizes of values that are inserted via continuous ingest. Zipfian distribution was selected since it tends to correspond to the distribution of real-world events.

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
@@ -38,6 +38,7 @@ import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.security.ColumnVisibility;
 import org.apache.accumulo.testing.TestProps;
 import org.apache.accumulo.testing.util.FastFormat;
+import org.apache.commons.math3.random.RandomDataGenerator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java
```diff
@@ -173,6 +181,18 @@ protected static void doIngest(AccumuloClient client, long rowMin, long rowMax,
     log.info("DELETES will occur with a probability of {}",
         String.format("%.02f", deleteProbability));
 
+    zipfianEnabled = Boolean.parseBoolean(testProps.getProperty("test.ci.ingest.zipfian.enabled"));
+
+    if (zipfianEnabled) {
+      minSize = Integer.parseInt(testProps.getProperty("test.ci.ingest.zipfian.min.size"));
+      maxSize = Integer.parseInt(testProps.getProperty("test.ci.ingest.zipfian.max.size"));
```
ctubbsii, line 188: Should these be `Long.parseLong()` instead or is Integer enough?

DomGarguilo: I think Integer is enough. The place where this is used (`rnd.nextZipf()`) only accepts an int anyways:
https://github.com/apache/accumulo-testing/blob/db3835e37f9ebcd336c9cc62578366539f15aa24/src/main/java/org/apache/accumulo/testing/continuous/ContinuousIngest.java#L349-L350

ctubbsii (approved):
I haven't tried running this, but I'm fine with merging this sooner rather than wait, if you want to get it in to test with it. It can always be improved later. The testing repo is really for us anyway, not something we release.

ctubbsii:
@DomGarguilo what's the status of this? It seems ready to merge. If you merge it to 2.1, you should subsequently merge it to 3.1 and then merge 3.1 to main.

DomGarguilo:
> @DomGarguilo what's the status of this? It seems ready to merge. If you merge it to 2.1, you should subsequently merge it to 3.1 and then merge 3.1 to main.

Yep, ready to merge. Thanks for switching the base branch. I'll merge it in and merge it up as suggested.

PR #277, opened 2024-04-15
Replace map reduce prop in continuous ingest verify

### LINKED ISSUE #274: mapred.map.tasks.speculative.execution is deprecated. Instead, use mapreduce.map.speculative
`INFO Configuration.deprecation: mapred.map.tasks.speculative.execution is deprecated. Instead, use mapreduce.map.speculative` gets logged when I run `cingest verify`.

This is set here:
https://github.com/apache/accumulo-testing/blob/ce95056c57d3eff61400486b4a9da8fd73dbe67e/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java#L193

It should be determined if the suggested replacement will work the way we expect and if it does, replace the deprecated field.

DomGarguilo:
Fixes #274

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java
@@ -190,7 +190,7 @@ public class ContinuousVerify extends Configured implements Tool {
 
       job.setOutputFormatClass(TextOutputFormat.class);
 
-      job.getConfiguration().setBoolean("mapred.map.tasks.speculative.execution", scanOffline);
+      job.getConfiguration().setBoolean("mapred.map.speculative", scanOffline);
       job.getConfiguration().set("mapreduce.job.classloader", "true");
 
       Path outputPath = new Path(outputDir + "/" + job.getJobName());


dlmarion (approved):
This property name looks correct for the version of Hadoop

ctubbsii:
When was this property changed in Hadoop? We need to be careful we're not breaking things with the earliest Hadoop version we still support.

DomGarguilo:
> When was this property changed in Hadoop? We need to be careful we're not breaking things with the earliest Hadoop version we still support.

Honestly I'm not sure. I am having trouble finding what version this change happened in. I do see that the following code was from 13 years ago at which point this property was already deprecated.

https://github.com/apache/hadoop/blame/833da65d8629e0bdd79097ed26033add9fc1864f/hadoop-mapreduce-project/hadoop-mapreduce-client/hadoop-mapreduce-client-core/src/main/java/org/apache/hadoop/mapreduce/util/ConfigUtil.java#L325

ctubbsii:
`git tag --contains 833da65d8629e0bdd79097ed26033add9fc1864f` shows that it appears in at least Hadoop 3.0.0 and newer, and we don't support anything older than that anymore, so this should be good.

ctubbsii (approved):

PR #279, opened 2024-07-03
Modified monitoring stack to use Telegraf and Timely

dlmarion:
This commit modifies the Terraform configuration such that instead of deploying a Telegraf/InfluxDB/Grafana container on the Manager node, it now deploys Telegraf on each node and uses a Timely/Grafana container on the Manager node. The processes running on each node will send metrics to the Telegraf process running locally,which will then forward the metrics to Timely to be inserted into Accumulo. Thus, as the cluster grows larger, more and more metrics will be created and inserted into Accumulo.


DomGarguilo:
With these changes, does the Deployment Overview section in the README need to be updated?

https://github.com/apache/accumulo-testing/blob/f191f7b23ebcf28388a33efcf71ed8437f1c8cbc/contrib/terraform-testing-infrastructure/README.md#deployment-overview

ctubbsii:
@dlmarion what's the status of this?

dlmarion:
@ctubbsii - status is I was just waiting for review to happen. I'll make sure that it's up date.

ctubbsii (commented):
I'm not opposed to merging this, but I'm not clear on its value either. I don't know how much value InfluxDB offered, vs. Timely.

"It's fine, I guess", is about all I can say, if people find it useful for testing Accumulo. I don't frequently use the Terraform stuff, though, and haven't tested this myself.

dlmarion:
> I'm not opposed to merging this, but I'm not clear on its value either.

The value here is that using Timely instead of InfluxDB places load on Accumulo as Timely stores its data in Accumulo. Having the cluster running inserts data without having to run CI.

PR #283, opened 2024-10-04
merge main into elasticity

DomGarguilo:
Merged main branch into elasticity branch and dropped `TABLE_MINC_COMPACT_IDLETIME` property use.


dlmarion (approved):
LGTM

ctubbsii:
It looks like this was squash merged, so the elasticity branch doesn't actually contain the history of the main branch. We need to be careful about that when trying to do a PR for a merge across branches.

PR #289, opened 2024-12-10
fixes sorting in bulk batch writer

keith-turner:
BulkBatchWriter was sorting only on the mutation row.  Occasionally continuous ingest would generate the same 64 bit id twice in the same batch and if its columns were not sorted properly then the write to the rfile would fail.

This fixes the problem by sorting on the entire key instead of only the row.


keith-turner:
Ran continuous ingest for a bit  and then verify to test these changes.

ctubbsii (approved):

PR #290, opened 2025-02-11
Replace call to ColumnVisibility.flatten() with getExpression()

DomGarguilo:
`ColumnVisibility.flatten()` has been removed. This PR replaces it with `ColumnVisibility.getExpression()`

### DIFF
--- a/src/main/java/org/apache/accumulo/testing/continuous/ContinuousInputFormat.java
+++ b/src/main/java/org/apache/accumulo/testing/continuous/ContinuousInputFormat.java
@@ -143,8 +143,7 @@ public class ContinuousInputFormat extends InputFormat<Key,Value> {
 
         byte[] fam = genCol(random.nextInt(maxFam));
         byte[] qual = genCol(random.nextInt(maxQual));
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/accumulo/testing/continuous/ContinuousInputFormat.java
```diff
@@ -143,8 +143,7 @@ private Key genKey(CRC32 cksum) {
 
         byte[] fam = genCol(random.nextInt(maxFam));
         byte[] qual = genCol(random.nextInt(maxQual));
-        @SuppressWarnings("deprecation")
-        byte[] cv = visibilities.get(random.nextInt(visibilities.size())).flatten();
+        byte[] cv = visibilities.get(random.nextInt(visibilities.size())).getExpression();
```
ddanielr, line 146: `flatten` was used to normalize the existing parse tree. 
While calling `getExpression()` will  return a byte array, I don't believe it de-duplicates the expression anymore.

Just something to be aware of when testing.

ddanielr (approved):

PR #305, opened 2026-03-12
changes to support building with java 21

keith-turner:


src/main/java/org/apache/accumulo/testing/stress/DataWriter.java
```diff
@@ -18,27 +18,19 @@
...
   private final RandomMutations mutations;
-  private final Cleanable cleanable;
   private final AtomicBoolean closed = new AtomicBoolean(false);
 
-  private static final Logger log = LoggerFactory.getLogger(DataWriter.class);
-
   public DataWriter(BatchWriter writer, RandomMutations mutations) {
     this.writer = writer;
     this.mutations = mutations;
-    this.cleanable = CleanerUtil.unclosed(this, DataWriter.class, closed, log, writer);
```
keith-turner, line 41: This was causing problems because `this` escaped in a constructor.  DataWriter is only used in try with resources, so this extra complexity for the cleaner seems unessecary.  Removed it instead of trying to fix the issue w/ this.

PR #309, opened 2026-06-02
Update pom and fix build

DomGarguilo:
This PR...
* aligns the pom with the main accumulo repo 
* replaces a deprecated var so the maven build succeeds


DomGarguilo:
610331b I needed to add the snapshots repo to the pom as outline here: https://maven.apache.org/pom.html#Repositories

failure: https://github.com/DomGarguilo/accumulo-testing/actions/runs/26836197625/job/79130627288

pom.xml
```diff
@@ -121,19 +120,32 @@
...
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+      <id>apache.snapshots</id>
+      <name>Apache Snapshot Repository</name>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
+    </repository>
+  </repositories>
```
DomGarguilo, line 135: I have opted to leave this block in as opposed to adding `-Dapache.snapshots` to the github actions maven command(s).

This block is necessary to pull from the snapshots repo. On a fresh install of accumulo-testing, its possible that the user does not have the current accumulo snapshot already in their `.m2`. Adding this repository block to the pom makes things more self contained and makes things work regardless of the state of the machine that is running things.

ctubbsii: @DomGarguilo The block still needs to be removed. It can be replaced with a property in the pom.xml that sets `<apache.snapshots>true</apache.snapshots>`

ctubbsii: However, it is very risky to leave it in place. Having it enabled by default across the board can get a user blocked by ASF. Enabling it requires some understanding of the risks, and should be a conscious choice. Users can set it in a maven profile or in their `MAVEN_OPTS` if they understand the risks and wish to enable it. Or, they can choose to take a safer option and build a local snapshot with `mvn install`, which should really be the default path, since it's the safest, and most useful (it tests the current snapshot, not whichever the last one that was published to the snapshot repo, which could be out of date).

ctubbsii: Fixed in 7e8ff9ee880b55dbed5e85d36a8eb3b920c0d16f

PR #314, opened 2026-09-11
Stop continuous ingest on flush boundary via ctrl+c

DomGarguilo:
When stopping continuous ingest via ctrl+c, the writes stop mid-chain. These changes add a shutdown hook that stops ingest cleanly after the next flush so that `cingest verify` won't report misleading unreferenced counts.


conf/log4j2.properties
```diff
@@ -21,6 +21,8 @@ name = AccumuloTestingDefaultLoggingProperties
 status = info
 dest = err
 monitorInterval = 30
+# let ctrl-c logging keep working while ingest finishes its current flush
+shutdownHook = disable
```
dlmarion, line 25: Is this used?

DomGarguilo: Yea. This makes it so that when we ctrl + c the last flush logs get printed. Without it, logging stops right away which makes it seems like the last flush did not happen.

dlmarion: I guess I'm missing where it's used in the code below. Could you point it out?

DomGarguilo: Its not referenced in our code. Its a log4j setting. `cingest` passes the following
https://github.com/apache/accumulo-testing/blob/4e1362821d5580a67fb889f4c182ce50869cb7b2/bin/cingest#L108

Which picks up the settings file touched in this PR.

dlmarion: Ok, I see now. Thanks for the explanation.

dlmarion (approved):

